PR #1, opened 2025-01-28
Rename `metatensor.torch.atomistic` to `metatomic.torch`

Luthaf:
Here we go!


PicoCentauri:
and so it begins... :-)

PR #37, opened 2025-05-26
Multiple improvements to ASE calculator

### LINKED ISSUE #8: Handle out of order samples in the ASE interface
When using per atom output, we don't guarantee that the results are in the same order as the atoms in the system. The calculator should handle this.

Luthaf:
See the commits for a description of the individual changes

Fix #8 

# Reviewer checklist

### DIFF
--- a/python/examples/4-profiling.py
+++ b/python/examples/4-profiling.py
@@ -183,7 +183,7 @@ print(energy_profiler.key_averages().table(sort_by="self_cpu_time_total", row_li
 # with the same function name as the corresponding torch functions (e.g.
 # ``aten::arange`` is :py:func:`torch.arange`). We can also see some internal functions
 # from metatomic, with the name staring with ``AtomisticModel::`` for
-# :py:class:`AtomisticModel`; and ``ASECalculator::`` for
+# :py:class:`AtomisticModel`; and ``MetatomicCalculator::`` for
 # :py:class:`ase_calculator.MetatomicCalculator`.
 #
 # If you want to see more details on the internal steps taken by your model, you can add
@@ -191,7 +191,7 @@ print(energy_profiler.key_averages().table(sort_by="self_cpu_time_total", row_li
 # (https://pytorch.org/docs/stable/generated/torch.autograd.profiler.record_function.html)
 # inside your model code to give names to different steps in the calculation. This is
 # how we are internally adding names such as ``Model::forward`` or
-# ``ASECalculator::prepare_inputs`` above.
+# ``MetatomicCalculator::prepare_inputs`` above.
 #
 
 # %%
--- a/python/examples/4-profiling.py
+++ b/python/examples/4-profiling.py
@@ -134,33 +134,37 @@ wrapper.export("exported-model.pt")
 #     </details>
 #
 
-
 # %%
 #
-# If you are trying to profile your own model, you can start here and create a
+# If you are trying to profile your own model, you can start here and create
 # ``MetatomicCalculator`` with your own model.
-
-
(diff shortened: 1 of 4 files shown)


ceriottm:
> > It'd be wonderful if this also provided a mode to activate non-conservative force calculation, seems well on topic
> 
> This is already available: https://docs.metatensor.org/metatomic/latest/torch/reference/ase.html#metatomic.torch.ase_calculator.MetatomicCalculator has a `non_conservative` flag

There is some weird interaction with non_conservative_stress. In practice, if you try to get_forces with the flag set, it'll complain about "KeyError: 'non_conservative_stress'"

python/metatomic_torch/metatomic/torch/ase_calculator.py
```diff
@@ -68,6 +68,7 @@ def __init__(
         check_consistency=False,
         device=None,
         non_conservative=False,
+        do_energy_gradients_with_energy=True,
```
Luthaf, line 71: I'm taking naming suggestions for this parameter!

HaoZeke: I kinda like `precompute_energy_gradients` but maybe `do_energy_gradients_with_energy` is the most descriptive..

python/examples/4-profiling.py
```diff
@@ -134,31 +134,36 @@ def forward(
...
 
 # %%
 #
 # Before trying to profile the code, it is a good idea to run it a couple of times to
 # allow torch to warmup internally.
 
-atoms.get_forces()
-atoms.get_potential_energy()
+for _ in range(10):
+    atoms.get_potential_energy()
```
HaoZeke, line 161: Do we not cache the energy on the `atoms` object making this a no-op after the first calculation?

python/metatomic_torch/metatomic/torch/ase_calculator.py
```diff
@@ -96,10 +88,17 @@ def __init__(
...
+            non-conservative forces and stresses. This can afford a speed-up,
             potentially at the expense of physical correctness (especially in molecular
             dynamics simulations).
+        :param do_energy_gradients_with_energy: if ``True``, this calculator will always
+            compute the energy gradients (forces and stress) when the energy is
+            requested (e.g. through ``atoms.get_potential_energy()``). Because the
+            results of a calculation are cached by ASE, this means future calls to
+            ``atom.get_forces()`` will return immediatly, without needing to execute the
+            model again. If you are mainly interested in the energy, you can set this to
+            ``False`` and enjoy a faster model.
```
HaoZeke, line 101:
```suggestion
            model again. If you are mainly interested in the energy, you can set this to
            ``False`` and enjoy a faster model. Forces will still be calculated and cached with if requested with `get_forces()`
```

python/examples/4-profiling.py
```diff
@@ -202,6 +207,12 @@ def forward(
 # Let's now do the same, but computing the forces for this system. This mean we should
 # now see some time spent in the ``backward()`` function, on top of everything else.
 
+atoms.calc = MetatomicCalculator("exported-model.pt")
+
+# warmup
+for _ in range(10):
+    atoms.get_forces()
```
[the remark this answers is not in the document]
Luthaf: This is true, but I would also prefer that people don't rely on it too much. This should just be an optimization

python/metatomic_torch/metatomic/torch/ase_calculator.py
ceriottm: `do_gradients_with_energy` it is

frostedoyster (approved):

PR #39, opened 2025-06-03
DOC: Fix typo

HaoZeke:
Minor nit.


docs/src/engines/lammps.rst
```diff
@@ -218,6 +218,7 @@ configure the build with:
 
     # you can add more options here to enable other packages.
     cmake -DPKG_ML-METATENSOR=ON \
```
Luthaf, line 220: the ML-METATENSOR=ON should also be removed

docs/src/engines/lammps.rst
```diff
@@ -227,6 +228,8 @@ configure the build with:
     # optionally install the code on your machine. You can also directly use
     # the `lmp` binary in `lammps-metatomic/build/lmp` without installation
     cmake --build . --target install # or `make install`
+    # For usage with lammps within ase
+    make install-python
```
Luthaf, line 232: With the configuration marked above, this will not work since we are not enabling the Python side of things. So I would rather not put anything here and refer people to the official documentation for anything not related to metatomic

HaoZeke: Yeap makes sense.

PR #40, opened 2025-06-04
Documentation for non-conservative and scaling factor in LAMMPS

frostedoyster:


docs/src/engines/lammps.rst
```diff
@@ -287,6 +287,14 @@ documentation.
         directory = path to a directory containing TorchScript extensions as
         shared libraries. If the model uses extensions, we will try to load
         them from this directory first
+      **non_conservative** values = on or off
+        set this to on to use non-conservative forces and stresses in your
+        simulation, affordint a speedup factor between 2 and 3. We recommend
+        using this in combination with RESPA to obtain physically correct
+        observables (see ...)
```
Luthaf, line 294: can you put the arxiv link there for now?

frostedoyster: In my mind this was going to be a recipe (because otherwise there is no chance that a user can do it themselves), but I guess we can also use arxiv. In any case, we have to wait for the recipe

frostedoyster: Actually the recipe will be the same as the one using i-PI, so we already have the link. I've added it

docs/src/engines/lammps.rst
```diff
@@ -287,6 +287,16 @@ documentation.
         directory = path to a directory containing TorchScript extensions as
         shared libraries. If the model uses extensions, we will try to load
         them from this directory first
+      **non_conservative** values = on or off
+        set this to on to use non-conservative forces and stresses in your
+        simulation, typically affording a speedup factor between 2 and 3. We recommend
+        using this in combination with RESPA to obtain physically correct
+        observables (see https://arxiv.org/abs/2412.11569 for more information, and
+        https://atomistic-cookbook.org/examples/pet-mad-nc/pet-mad-nc.html for an
+        example of how to set up the RESPA run)
```
Luthaf, line 296:
```suggestion
        example of how to set up the RESPA run). Default to off.
```

PR #52, opened 2025-07-07
[DOC] Add eOn engine details

HaoZeke:
Just adds some minimal documentation for the new eOn backend.

The big question is what would be an interesting example here, the simplest one is just single point and then in complexity there's minimization, saddle searches (dimer / NEB / GP-dimer) or full aKMC but I don't really think those should be here instead of in the cookbook.

### DIFF
--- a/docs/src/engines/eon.rst
+++ b/docs/src/engines/eon.rst
@@ -30,11 +30,9 @@ well.
 How to install the code
 ^^^^^^^^^^^^^^^^^^^^^^^
 
-The metatomic interface is part of eOn was completed in `PR-206`_. Please refer
-to latest `eOn documentation`_ about how to install it.
+Please refer to latest `eOn documentation`_ about how to install it.
 
 .. _eOn documentation: https://theochemui.github.io/eOn/install/metatomic.html
-.. _PR-206: https://github.com/TheochemUI/eOn/pull/206
 
 How to use the code
 ^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


docs/src/engines/eon.rst
```diff
@@ -0,0 +1,66 @@
...
+well.
+
+How to install the code
+^^^^^^^^^^^^^^^^^^^^^^^
+
+The metatomic interface is part of eOn was completed in `PR-206`_. Please refer
+to latest `eOn documentation`_ about how to install it.
+
+.. _eOn documentation: https://theochemui.github.io/eOn/install/metatomic.html
+.. _PR-206: https://github.com/TheochemUI/eOn/pull/206
```
Luthaf, line 37:
```suggestion
Please refer to latest `eOn documentation`_ about how to install it.

.. _eOn documentation: https://theochemui.github.io/eOn/install/metatomic.html
```

I don't think users will care about the PR that added the code =). Or do you want to say "this is not released yet, please see the instructions to build from source"?

docs/src/engines/eon.rst
```diff
@@ -0,0 +1,66 @@
...
+
+.. code-block:: ini
+
+    [Potential]
+    potential = metatomic
+
+    [Metatomic]
+    model_path = # $FULL_PATH/pet-mad-full-best.pt
+
+Where it is more robust to use the complete model path, especially for the adaptive kinetic monte carlo runs. Complete details of the input file specification are present in the `upstream reference documentation`_.
```
Luthaf, line 64:
```suggestion
Where it is more robust to use the complete model path, especially for the adaptive kinetic monte carlo runs. Complete details of the input file specification are present in the `corresponding reference documentation`_.
```

docs/src/engines/eon.rst
```diff
@@ -0,0 +1,66 @@
...
+
+.. _upstream reference documentation: https://theochemui.github.io/eOn/user_guide/index.html
```
Luthaf, line 66:
```suggestion
.. _corresponding reference documentation: https://theochemui.github.io/eOn/user_guide/index.html
```

Luthaf:
> The big question is what would be an interesting example here, the simplest one is just single point and then in complexity there's minimization, saddle searches (dimer / NEB / GP-dimer) or full aKMC but I don't really think those should be here instead of in the cookbook.

I'd say what's already here is fine, and examples can be added to the cookbook. We could add links in metatomic docs to the cookbook examples though!

Luthaf (approved):

Issue #54, opened 2025-07-12
Expose ase_calculator


ceriottm:
Currently if one wants to use both `ase_calculator.MetatomicCalculator` and say ModelOutput one has to import separately MTA and the calculator, e.g.

```
import metatomic.torch as mta
from metatomic.torch.ase_calculator import MetatomicCalculator
```

because `MetatomicCalculator` is not re-exported, nor the `ase_calculator` module. 

Can we expose either ase_calculator or (maybe better) directly `MetatomicCalculator` from the MTA module?

ceriottm:
can we at least do so that import mta allows to access mta.ase? It's really annoying to have to import the two separately.

Luthaf:
Yes, this might be possible with a [module-level `__getattr__`](https://peps.python.org/pep-0562/) that would lazily import the code. I'll have a look.

PR #60, opened 2025-07-14
[MAINT] Better lazy loading

HaoZeke:
Reworks the fix to #54 a bit by:
- Caching the module to prevent multiple imports
  - By vendoring a stripped version of [lazy-loader](https://github.com/scientific-python/lazy-loader/tree/main)
- Adding tests
- Demonstrating lazy re-exports of both `ase_calculator` and `MetatomicCalculator`.

~**Do not merge** until a consensus of which to export is reached (both, as shown in this PR is a bit redundant but.. also OK).~ Both is fine.

 - [X] Tests updated (for new features and bugfixes)?
 - [X] Issue referenced (for PRs that solve an issue)?

# Reviewer checklist

### DIFF
--- a/python/metatomic_torch/metatomic/torch/__init__.py
+++ b/python/metatomic_torch/metatomic/torch/__init__.py
@@ -59,7 +59,5 @@ def __getattr__(name):
     if name in ("ase_calculator", "MetatomicCalculator"):
         if name == "ase_calculator":
             return importlib.import_module(".ase_calculator", __name__)
-        else:  # name == "MetatomicCalculator"
-            return importlib.import_module(".ase_calculator", __name__).MetatomicCalculator
 
     raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
--- a/python/metatomic_torch/metatomic/torch/__init__.py
+++ b/python/metatomic_torch/metatomic/torch/__init__.py
@@ -1,9 +1,11 @@
+import importlib
 import os
 
 import torch
 
 from ._c_lib import _load_library
 from .version import __version__  # noqa: F401
+from . import utils
 
 
 if os.environ.get("METATOMIC_IMPORT_FOR_SPHINX", "0") != "0":
@@ -50,12 +52,20 @@ from .systems_to_torch import systems_to_torch  # noqa: F401
 
 
 def __getattr__(name):
-    # lazy import for ase_calculator, making it accessible as
-    # ``metatomic.torch.ase_calculator`` without requiring a separate import from
-    # ``metatomic.torch``, but only importing the code when actually required.
-    if name == "ase_calculator":
-        import metatomic.torch.ase_calculator
+    """
+    Lazily import the `ase_calculator` module or the `MetatomicCalculator`
+    class upon first access.
+    """
+    if name in ("ase_calculator", "MetatomicCalculator"):
+        # Import the submodule once using importlib for a clean, relative import
+        module = importlib.import_module(".ase_calculator", __name__)
 
-        return metatomic.torch.ase_calculator
-    else:
-        raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+        if name == "ase_calculator":
+            globals()[name] = module
+            return module
+        else:  # name == "MetatomicCalculator"
+            calculator_class = module.MetatomicCalculator
+            globals()[name] = calculator_class
+            return calculator_class
+
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
--- a/python/metatomic_torch/metatomic/torch/__init__.py
+++ b/python/metatomic_torch/metatomic/torch/__init__.py
@@ -3,9 +3,9 @@ import os
 
 import torch
 
+from . import utils  # noqa: F401
 from ._c_lib import _load_library
 from .version import __version__  # noqa: F401
-from . import utils
 
 
 if os.environ.get("METATOMIC_IMPORT_FOR_SPHINX", "0") != "0":
--- a/python/metatomic_torch/metatomic/torch/__init__.py
+++ b/python/metatomic_torch/metatomic/torch/__init__.py
@@ -1,5 +1,6 @@
 import importlib
 import os
+import sys
 
 import torch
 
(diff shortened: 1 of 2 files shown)


python/metatomic_torch/tests/ase_calculator.py
```diff
@@ -549,3 +551,73 @@ def test_additional_outputs(atoms):
...
+    Closes gh-54.
+    """
+    import metatomic.torch
+
+    module_name = "metatomic.torch.ase_calculator"
+
+    assert module_name not in sys.modules
+    ase_module = metatomic.torch.ase_calculator
+    assert module_name in sys.modules
+    assert isinstance(ase_module, types.ModuleType)
```
HaoZeke, line 604: Probably shouldn't keep this. Just `MetatomicCalculator` is likely enough..

python/metatomic_torch/tests/ase_calculator.py
```diff
@@ -549,3 +551,73 @@ def test_additional_outputs(atoms):
 
     another = atoms.calc.additional_outputs["another::one"]
     assert another.block().properties.names == ["another"]
+
+
+def with_isolated_imports(test_func):
```
HaoZeke, line 556: IMO not great to have a decorator here instead of in `configtest` but it also has zero use anywhere else.

python/metatomic_torch/metatomic/torch/__init__.py
```diff
@@ -50,12 +52,20 @@
...
+    """
+    if name in ("ase_calculator", "MetatomicCalculator"):
+        # Import the submodule once using importlib for a clean, relative import
+        module = importlib.import_module(".ase_calculator", __name__)
+
+        if name == "ase_calculator":
+            globals()[name] = module
+            return module
+        else:  # name == "MetatomicCalculator"
+            calculator_class = module.MetatomicCalculator
```
Luthaf, line 67: I don't think we want such a generic name as `mta.MetatomicCalculator` for ASE specific code. It should either be scoped like `mta.ase.MetatomicCalculator` or renamed `mta.MetatomicAseCalculator`.

python/metatomic_torch/metatomic/torch/__init__.py
```diff
@@ -50,12 +52,20 @@
...
+    Lazily import the `ase_calculator` module or the `MetatomicCalculator`
+    class upon first access.
+    """
+    if name in ("ase_calculator", "MetatomicCalculator"):
+        # Import the submodule once using importlib for a clean, relative import
+        module = importlib.import_module(".ase_calculator", __name__)
+
+        if name == "ase_calculator":
+            globals()[name] = module
+            return module
```
Luthaf, line 65: I don't get this. `import metatomic.torch.ase_calculator` should already cache modules in `sys.modules`. Why is this needed?

python/metatomic_torch/metatomic/torch/__init__.py
```diff
@@ -49,13 +52,63 @@
...
+# [2]: https://pypi.org/project/lazy-loader/
+# The key is the name exposed to the user.
+# The value is a tuple: (relative_module_path, attribute_name_in_module)
+# If attribute_name_in_module is None, the entire module is returned.
+_lazy_imports = {
+    # Exposes `metatomic.torch.ase_calculator` as `metatomic.torch.ase`
+    "ase": (".ase_calculator", None),
+    # Exposes `MetatomicCalculator` from the submodule as `MetatomicAseCalculator`
+    "MetatomicAseCalculator": (".ase_calculator", "MetatomicCalculator"),
+}
```
Luthaf, line 68: I don't think we want both `metatomic.torch.ase_calculator` and `metatomic.torch.ase` (and same for `metatomic.torch.ase_calculator.MetatomicCalculator` and `metatomic.torch.MetatomicAseCalculator`), it is going to be confusing to users and to us when debugging errors.

python/metatomic_torch/tests/ase_calculator.py
```diff
@@ -549,3 +550,22 @@ def test_additional_outputs(atoms):
...
+
+
+def test_lazy_loading_module():
+    """Tests that the `ase_calculator` module is lazy-loaded correctly."""
+    import metatomic.torch
+
+    module_name = "metatomic.torch.ase_calculator"
+    ase_module = metatomic.torch.ase
+    assert module_name in sys.modules
+    assert isinstance(ase_module, types.ModuleType)
```
Luthaf, line 562: Wouldn't this work just because other parts of this file are importing this code? It might be worth moving this to a separate file for testing.

Luthaf (commented):
I still don't understand what this is trying to solve 😅. Could you explain what the problem is and why it needs to be solved? 

I do see a point for a test, but I don't get why the mechanism needs to get more complex than just `import` or why we would want to rename/re-export under a different name.

Luthaf:
> This has the advantage of providing support for:

The code below is the same thing twice, was this intentional? 

> requiring the user to know what they're looking for.

Well, we do have all of it visible in the API reference and the examples, but we could also add it to `dir()`. This could also be done just by adding `ase_calculator` to a module-level `__dir__` function.

-------

Regarding the other changes in this PR, exposing `MetatomicCalculator` directly on the `mta` module would be a breaking change since we would also have to rename the class. I'm not sure the small added convenience is really worth it … And similar for renaming the module from `ase_calculator` to `ase`.

And none of these changes require changing how we do lazy loading under the hood!

Overall, I think we could add the tests and custom `__dir__` function now, but renaming classes/modules is not really worth it, and is something we should instead do when moving the calculator out of `metatomic.torch` and into `metatomic` as we make the code no longer reliant on TorchScript.

HaoZeke:
> > This has the advantage of providing support for:
> 
> The code below is the same thing twice, was this intentional?
> 
> > requiring the user to know what they're looking for.
> 
> Well, we do have all of it visible in the API reference and the examples, but we could also add it to `dir()`. This could also be done just by adding `ase_calculator` to a module-level `__dir__` function.
> 
> Regarding the other changes in this PR, exposing `MetatomicCalculator` directly on the `mta` module would be a breaking change since we would also have to rename the class. I'm not sure the small added convenience is really worth it … And similar for renaming the module from `ase_calculator` to `ase`.
> 
> And none of these changes require changing how we do lazy loading under the hood!
> 
> Overall, I think we could add the tests and custom `__dir__` function now, but renaming classes/modules is not really worth it, and is something we should instead do when moving the calculator out of `metatomic.torch` and into `metatomic` as we make the code no longer reliant on TorchScript.


Fair, I can rework it with just those. However..

> Regarding the other changes in this PR, exposing `MetatomicCalculator` directly on the `mta` module would be a breaking change since we would also have to rename the class. I'm not sure the small added convenience is really worth it … And similar for renaming the module from `ase_calculator` to `ase`.

With the current lazy loading schema there's no need to rename it, as it is available through either name.

PR #90, opened 2025-09-17
Add function to pick_device based on models availability and use request.

### LINKED ISSUE #21: Utility function to select the best device for a model
This operation requires a considerable amount of lines and has many pitfalls (e.g. `"cuda:2"`). It needs to be replicated for every supported engine.

PicoCentauri:
Fixes #21 

[We have something similar in metatrain
](https://github.com/metatensor/metatrain/blob/c07d491fa655862638c54e96e28b967f18eaa9f7/src/metatrain/utils/devices.py#L12)

Biggest question now is, how we handle multi GPU requests. We have this in LAMMPS and also would need this to replace the metatrain function. We could add another parameter?

# Reviewer checklist

[📚 Download documentation for this pull-request](https://nightly.link/metatensor/metatomic/actions/artifacts/4056553017.zip)

### DIFF
--- a/metatomic-torch/include/metatomic/torch/misc.hpp
+++ b/metatomic-torch/include/metatomic/torch/misc.hpp
@@ -3,12 +3,22 @@
 
 #include <string>
 
+#include <torch/types.h>
+
 #include "metatomic/torch/exports.h"
 
 namespace metatomic_torch {
 
 /// Get the runtime version of metatensor-torch as a string
 METATOMIC_TORCH_EXPORT std::string version();
+
+/// Select the best device according to the list of `model_devices` from a
+/// model, the user-provided `desired_device` and what's available on the 
+/// current machine.
+METATOMIC_TORCH_EXPORT std::string pick_device(
+	std::vector<std::string> model_devices,
+	torch::optional<std::string> desired_device = torch::nullopt
+);
 
 }
 
--- a/metatomic-torch/src/misc.cpp
+++ b/metatomic-torch/src/misc.cpp
@@ -1,6 +1,62 @@
+#include <torch/torch.h>
+
 #include "metatomic/torch/version.h"
 #include "metatomic/torch/misc.hpp"
 
 std::string metatomic_torch::version() {
     return METATOMIC_TORCH_VERSION;
+}
+
+std::string metatomic_torch::pick_device(
+    std::vector<std::string> model_devices,
+    torch::optional<std::string> desired_device
+) {
+    auto available_devices = std::vector<std::string>();
+    std::string selected_device = "cpu";
+
+    for (const auto& device: model_devices) {
+        if (device == "cpu") {
+            available_devices.push_back("cpu");
+        } else if (device == "cuda") {
+            if (torch::cuda::is_available()) {
+                available_devices.push_back("cuda");
+            }
+        } else if (device == "mps") {
+            if (torch::mps::is_available()) {
+                available_devices.push_back("mps");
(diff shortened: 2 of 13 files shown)


metatomic-torch/src/register.cpp
```diff
@@ -207,6 +207,7 @@ TORCH_LIBRARY(metatomic, m) {
 
     // standalone functions
     m.def("version() -> str", metatomic_torch::version);
+    m.def("pick_device(list supported_devices, str requested_device = None) -> torch.Device", metatomic_torch::pick_device);
```
Luthaf, line 210: `str requested_device = None` feels wrong, should it be `Optional[str] requested_device = None` or `str requested_device? = None` (not sure about the syntax for optional here)

metatomic-torch/include/metatomic/torch/misc.hpp
```diff
@@ -3,13 +3,22 @@
...
 namespace metatomic_torch {
 
+
 /// Get the runtime version of metatensor-torch as a string
 METATOMIC_TORCH_EXPORT std::string version();
 
+/// Select device on 'supported_devices', a 'desired_device' and what's available on the current machine.
+METATOMIC_TORCH_EXPORT torch::Device pick_device(
+	const std::vector<std::string>& supported_devices,
+	const std::string& desired_device = ""
```
Luthaf, line 19:
```suggestion
	torch::optional<std::string> desired_device = torch::nullopt
```

This should be a bit nicer

Luthaf:
> Biggest question now is, how we handle multi GPU requests. We have this in LAMMPS and also would need this to replace the metatrain function. We could add another parameter?

Either we handle it here or inside LAMMPS. Maybe doing it inside LAMMPS makes the most sense since that's where we know about more things (in a sense this function is only picking the device type, not the device id).

```
device = pick_best_device(...)
if device.is_cuda:
     # change the device ID with the current code
````

File: metatomic-torch/src/misc.cpp [hunk 3]
```diff
@@ -1,6 +1,66 @@
...
+    const std::string& desired_device) {
+
+    auto available_devices = std::vector<torch::Device>();
+    auto selected_device = torch::Device(torch::kCPU);
+
+    for (const auto& device: supported_devices) {
+        if (device == "cpu") {
+            available_devices.push_back(torch::Device(torch::kCPU));
+        } else if (device == "cuda" && torch::cuda::is_available()) {
+            available_devices.push_back(torch::Device(torch::kCUDA, 1));
```
Luthaf, line 21:
```suggestion
            available_devices.push_back(torch::Device(torch::kCUDA, 0));
```

metatomic-torch/include/metatomic/torch/misc.hpp
```diff
@@ -3,13 +3,22 @@
...
 #include "metatomic/torch/exports.h"
 
 namespace metatomic_torch {
 
+
 /// Get the runtime version of metatensor-torch as a string
 METATOMIC_TORCH_EXPORT std::string version();
 
+/// Select device on 'supported_devices', a 'desired_device' and what's available on the current machine.
+METATOMIC_TORCH_EXPORT torch::Device pick_device(
```
Luthaf, line 17:
```suggestion
/// Select the best device according to the list of `supported_devices` from a
/// model, the user-provided `desired_device` and what's available on the 
/// current machine.
METATOMIC_TORCH_EXPORT torch::Device pick_best_device(
```

PR #93, opened 2025-09-30
Use clang-format to format C++ code

frostedoyster:
As in the title


Luthaf:
Unless you have strong feeling in favor of doing this, I would rather keep the code manually formatted for now. I find that clang-format can create relatively hard to read code, and I'm happy going back over contributed code to clean it up if needed.

Luthaf:
Then let's not do that for now, we can revisit later

Issue #104, opened 2025-10-14
Implement a wrapper for heat flux calculation


GardevoirX:
The idea is to implement the heat flux calculation as a torch script, and as a result it can be easily used in both ASE and LAMMPS.

The model will be implemented in this way (Thanks @Luthaf ):

```python
class HeatFlux(torch.nn.Module):

    def __init__(self, model: AtomisticModel):
        super().__init__()
        self.model = model.module

    def forward(
        self, 
        systems: List[System], 
        outputs: Dict[str, ModelOutput], 
        selected_atoms = None
    ) -> Dict[str, TensorMap]:
        
        if "experimental::heat_flux" in outputs:
            ..

        else:
            return self.model(systems, outputs, selected_atoms)



wrapper = HeatFlux(pet_model)
wrapper.save("heat_flux.pt")

calc = MetatomicCalculator("heat_flux.pt")
outputs = calc.run_model(atoms, {"experimental::heat_flux": ModelOutput(per_atom=False)})

heat_flux = outputs["experimental::heat_flux"].block().values
```

And will be located in the `models` folder under `metatomic/python/metatomic_torch/metatomic/torch/`.

PR #111, opened 2025-11-02
Replace the current `_check_outputs` with its cpp version

GardevoirX:
The `_check_outputs` function is moved to cpp in order to meet the new requirements brought by #104 , in which we want to reuse the class `ModelOutput` to communicate with the simulation engine, and thus need a check function for that.

# Reviewer checklist

### DIFF
--- a/metatomic-torch/include/metatomic/torch/outputs.hpp
+++ b/metatomic-torch/include/metatomic/torch/outputs.hpp
@@ -11,17 +11,21 @@
 #include "metatomic/torch/model.hpp"
 #include "metatomic/torch/system.hpp"
 
-using namespace metatomic_torch;
-/// Check that the outputs of a model conform to the expected structure for atomistic
-/// models.
+namespace metatomic_torch {
 
+/// Check that the outputs of a model conform to the expected structure for
+/// atomistic models.
+///
 /// This function checks conformance with the reference documentation in
 /// https://docs.metatensor.org/metatomic/latest/outputs/index.html
-void _check_outputs(
+void METATOMIC_TORCH_EXPORT check_outputs(
     const std::vector<System>& systems,
     const c10::Dict<std::string, ModelOutput>& requested,
     const torch::optional<metatensor_torch::Labels>& selected_atoms,
     const c10::Dict<std::string, metatensor_torch::TensorMap>& outputs,
-    const int64_t dtype);
+    std::string model_dtype
+);
+
+}
 
 #endif
--- a/metatomic-torch/src/outputs.cpp
+++ b/metatomic-torch/src/outputs.cpp
@@ -1,29 +1,25 @@
-#include <torch/script.h>
-
 #include <algorithm>
 #include <cassert>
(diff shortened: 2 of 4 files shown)


metatomic-torch/include/metatomic/torch/outputs.hpp
```diff
@@ -0,0 +1,28 @@
...
+    const c10::Dict<std::string, ModelOutput>& requested,
+    const std::optional<metatensor_torch::Labels>& selected_atoms,
+    const c10::Dict<std::string, metatensor_torch::TensorMap>& outputs,
+    const torch::Dtype& expected_dtype);
+}
+    // Check that the outputs of a model conform to the expected structure for atomistic
+    // models.
+
+    // This function checks conformance with the reference documentation in
+    // https://docs.metatensor.org/metatomic/latest/outputs/index.html
```
Luthaf:
```suggestion
/// Check that the outputs of a model conform to the expected structure for atomistic
/// models.
///
/// This function checks conformance with the reference documentation in
/// https://docs.metatensor.org/metatomic/latest/outputs/index.html
void _check_outputs(
    const std::vector<System>& systems,
    const c10::Dict<std::string, ModelOutput>& requested,
    const std::optional<metatensor_torch::Labels>& selected_atoms,
    const c10::Dict<std::string, metatensor_torch::TensorMap>& outputs,
    const torch::Dtype& expected_dtype);
}
```

Documentation goes before a function in C++

metatomic-torch/src/outputs.cpp
```diff
@@ -0,0 +1,548 @@
...
+
+#include "./internal/utils.hpp"
+#include "metatomic/torch/exports.h"
+#include "metatomic/torch/model.hpp"
+#include "metatomic/torch/system.hpp"
+
+namespace metatomic_torch {
+
+const std::array energy_bases = {"energy", "energy_ensemble", "energy_uncertainty"};
+const std::array energy_gradients = {"strain", "positions"};
```
Luthaf: Please use upper case for global variables, and make them `static`

metatomic-torch/src/outputs.cpp
```diff
@@ -0,0 +1,548 @@
...
+void _validate_atomic_samples(
+    const std::string& name,
+    const metatensor_torch::TensorMap& value,
+    const std::vector<System>& systems,
+    const ModelOutput& request,
+    const torch::optional<metatensor_torch::Labels>& selected_atoms) {
+    // Validates the sample labels in the output against the expected structure
+
+    const torch::Device& device = value->device();
+    const metatensor_torch::TensorBlock& block = value->block_by_id(value, 0);
```
Luthaf: `block_by_id` is a static method in C++, and should be called as `TensorMap::block_by_id(value, 0)`

Luthaf: (they are static because they need to take `torch::intrusive_ptr<TensorMapHolder>` and not TensorMapHolder as `this`).

PR #113, opened 2025-11-03
chore(devicePicker): handle more and return device type

HaoZeke:
I really like https://github.com/metatensor/metatomic/pull/90 however I realized in https://github.com/TheochemUI/eOn/pull/262 that it might be more ergonomic to pass around the `devicetype` instead of the string representation. The Python side should still get the `str` of course.

Plus this adds the rest of the devices supported [by Torch](https://github.com/pytorch/pytorch/blob/3eddf049221fc04c2ac9d4af53c00305484ef325/c10/core/Device.cpp#L17), except the `privateuse` thing which sounds kinda scammy.

# Reviewer checklist

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -31,7 +31,7 @@ test_options =
     --import-mode=append
 
 packaging_deps =
-    setuptools >= 75
+    setuptools >= 77
     packaging >= 23
     cmake
 


metatomic-torch/src/register.cpp
```diff
@@ -219,7 +219,27 @@ TORCH_LIBRARY(metatomic, m) {
...
+            try {
+                torch::optional<std::string> desired = torch::nullopt;
+                if (requested_device.has_value()) {
+                    desired = requested_device.value();
+                }
+                c10::DeviceType devtype = metatomic_torch::pick_device(model_devices, desired);
+                torch::Device dev(devtype);
+                return dev.str();
+            } catch (const c10::Error &e) {
+                throw std::runtime_error(std::string("pick_device failed: ") + e.what());
```
PicoCentauri: Shouldn't this be a c10 error to be well forwarded to python

PicoCentauri (commented):
Thanks! I am fine with having a different API for C and Python. 

Maybe we also ask for a comment by @Luthaf.

Luthaf:
I'm happy to add support for all other device types in the function checking for availability, but I don't see a point returning a DeviceType instead of a string. This would be a breaking change requiring new releases of everything that depends on metatomic, and downstream code can always create the torch Device themself.

metatomic-torch/src/register.cpp
```diff
@@ -219,7 +219,32 @@ TORCH_LIBRARY(metatomic, m) {
 
     // standalone functions
     m.def("version() -> str", version);
-    m.def("pick_device(str[] model_devices, str? requested_device = None) -> str", pick_device);
+    // Expose pick_device to Python. The C++ helper returns a c10::DeviceType;
+    // build a torch::Device from it (with default index policy) and return its
+    // string representation to Python (backwards-compatible).
+    m.def(
+        "pick_device(str[] model_devices, str? requested_device = None) -> str",
+        [](const std::vector<std::string> &model_devices,
```
Luthaf: can you move this code in a standalone function on top of the file, to keep the bindings more readable?

Luthaf (commented):
What's the plan for backward compatibility? This is only breaking the C++ API, which should only be used in LAMMPS.

PicoCentauri:
PLUMED not using it
EON - does it
GROMACS - we are not using yet

HaoZeke:
Will merge when green ^_^

PR #148, opened 2026-02-02
feat(docs): add complete C++ API reference

HaoZeke:
Try them out here: https://github.com/metatensor/metatomic/actions/runs/21580495763/artifacts/5339250994

Before:
image

Which may bring up more questions than answers in terms of the actual implementation.

Now also has:
image
image

To be fair it's a bit meh here since these are basically empty, but still nice to see IMO.

FWIW this is basically "imagine if `showcode` in the Python worked for the C++ bits" 

# Reviewer checklist

### DIFF
--- a/scripts/mkdoc.nu
+++ b/scripts/mkdoc.nu
@@ -1,0 +1,54 @@
+#!/usr/bin/env nu
+
+def main [] {
+    let root = ($env.PWD)
+    let deps_dir = ($root | path join "docs/subprojects")
+    let doxyrest_src = ($deps_dir | path join "doxyrest")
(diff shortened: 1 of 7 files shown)


scripts/mkdoc.nu
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env nu
```
Luthaf: Sorry, we are not adding a new tool to the doc stack. Can this be made to work inside sphinx/tox?

Luthaf:
We can revisit this later if someone wants to pick it up

PR #152, opened 2026-02-04
Fix the default velocity unit of `ase`

GardevoirX:
The default unit is actually `(eV/u)^(1/2)` but I wrongly set it as `nm/fs`.

For reference, see https://ase-lib.org/ase/units.html, and also their implementation of `_maxwellboltzmanndistribution` https://gitlab.jsc.fz-juelich.de/kesselheim1/ase/-/blob/e240cec52abeb150b1b5cefa69888b7b60489878/ase/md/velocitydistribution.py#L14-21

# Reviewer checklist

### DIFF
--- a/python/metatomic_torch/tests/ase_calculator.py
+++ b/python/metatomic_torch/tests/ase_calculator.py
@@ -866,6 +866,8 @@ def test_additional_input(atoms):
 
         expected = ARRAY_QUANTITIES[name]["getter"](atoms).reshape(values.shape)
         if name == "velocities":
-            expected *= 0.098226948  # ase velocity is in (eV/u)^(1/2)
+            expected /= (
+                ase.units.Angstrom / ase.units.fs
+            )  # ase velocity is in (eV/u)^(1/2) and we want A/fs
 
         assert np.allclose(values, expected)


Luthaf (commented):
could you also check the unit for momenta?

GardevoirX:
> could you also check the unit for momenta?

yeah this one is correct, bascially `atoms.get_momenta() / atoms.get_velocities()` gives me the atomic mass in `u`

metatomic-torch/src/model.cpp
```diff
@@ -1141,6 +1141,7 @@ static std::map<std::string, Quantity> KNOWN_QUANTITIES = {
         {"A/fs", 1e1},
         {"m/s", 1e6},
         {"nm/ps", 1e3},
+        {"(eV/u)^(1/2)", 101.80506},
```
HaoZeke: Dunno if this is the right place but why not

(eV/u) **0.5

Do we recognize that too? I can't remember

HaoZeke: I mean... we could just do an expression parser or something, there's gotta be something out there already, with `sympy` and even `openmm` taking expression strings.

The point is rather than explicitly enumerating it'd be easier to just define the operators and be done.

GardevoirX: Yeah I also thought about this a little bit, if this is possible, we can utilize the `unit` stored in the `info` of the `TensorMap` and do unit calculations when functions in `metatensor.operations` are called, which makes our tensor more tensor-like. but maybe it's just nobody want to spend time on this

Maybe opening an issue on it is the best action at this moment

PR #199, opened 2026-04-03
Deprecate ModelOutput.quantity

HaoZeke:
The quantity field is no longer required for unit conversion since the unit parser determines dimensions from the expression itself.

Changes:
- Add deprecated documentation to C++ header
- Add TORCH_WARN in C++ setter when quantity is non-empty
- Add DeprecationWarning in Python when quantity validation runs
- Update documentation with deprecation notice
- Add deprecation entries to CHANGELOG

# Reviewer checklist

### DIFF
--- a/metatomic-torch/include/metatomic/torch/model.hpp
+++ b/metatomic-torch/include/metatomic/torch/model.hpp
@@ -84,13 +84,17 @@ public:
     /// description of this output, defaults to empty string of not set by the user
     std::string description;
 
-    /// quantity of the output (e.g. energy, dipole, …).  If this is an empty
+    /// quantity of the output (e.g. energy, dipole, ...).  If this is an empty
     /// string, no unit conversion will be performed.
+    /// @deprecated This field is no longer required for unit conversion.
+    ///             The unit parser determines dimensions from the expression itself.
+    [[deprecated("quantity is no longer required for unit conversion, use unit directly")]]
     const std::string& quantity() const {
         return quantity_;
     }
 
     /// set the quantity of the output
+    [[deprecated("quantity is no longer required for unit conversion, use unit directly")]]
     void set_quantity(std::string quantity);
 
     /// unit of the output. If this is an empty string, no unit conversion will
@@ -249,6 +253,10 @@ public:
     void set_length_unit(std::string unit);
 
     /// requested outputs for this run and corresponding settings
+
+    // FIXME: it would be nice to also check that the units are properly set
+    // here, but since this is a field and not a `set_outputs` function it will
+    // be hard to do, so we will wait until the next set of breaking changes.
     torch::Dict<std::string, ModelOutput> outputs;
 
     /// Only run the calculation for a selected subset of atoms. If this is set
--- a/metatomic-torch/src/model.cpp
+++ b/metatomic-torch/src/model.cpp
@@ -12,6 +12,7 @@
 
 #include "metatomic/torch/model.hpp"
 #include "metatomic/torch/misc.hpp"
+#include "metatomic/torch/outputs.hpp"
 #include "metatomic/torch/units.hpp"
 
 #include "./internal/shared_libraries.hpp"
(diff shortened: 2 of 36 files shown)


metatomic-torch/include/metatomic/torch/model.hpp
```diff
@@ -57,6 +57,8 @@ class METATOMIC_TORCH_EXPORT ModelOutputHolder: public torch::CustomClassHolder
 
     /// quantity of the output (e.g. energy, dipole, …).  If this is an empty
     /// string, no unit conversion will be performed.
+    /// @deprecated This field is no longer required for unit conversion.
+    ///             The unit parser determines dimensions from the expression itself.
```
Luthaf, line 90: Can you do a `[[deprecated]]` atribute instead, both here and on the setter?

metatomic-torch/src/model.cpp
```diff
@@ -54,6 +54,13 @@ static void read_vector_int_json(
 /******************************************************************************/
 
 void ModelOutputHolder::set_quantity(std::string quantity) {
+    if (!quantity.empty()) {
+        TORCH_WARN(
```
Luthaf: Should be a torch deprecation warning, there is a special macro for it

python/metatomic_ase/tests/conftest.py
```diff
@@ -6,5 +8,17 @@ def fail_test_with_output(capfd):
     yield
     captured = capfd.readouterr()
     # the code should not print anything to stdout or stderr
+    # except for expected deprecation warnings
```
Luthaf: These should be explicitly caught at the place where they are emitted in the test, by using the capfd fixture in the tests. I'm doing this in a couple of place already

python/metatomic_torch/metatomic/torch/model.py
```diff
@@ -510,6 +510,11 @@ def forward(
                 if declared.quantity == "" or requested.quantity == "":
                     continue
 
+                # Note: Deprecation warning for quantity is emitted from C++ when
+                # ModelOutput is constructed with non-empty quantity.
+                # We don't warn here because TorchScript can't handle
+                # DeprecationWarning.
+
```
Luthaf: we should remove the code here that uses `quantity`

python/metatomic_torch/tests/conftest.py
```diff
@@ -6,5 +8,17 @@ def fail_test_with_output(capfd):
     yield
     captured = capfd.readouterr()
     # the code should not print anything to stdout or stderr
+    # except for expected deprecation warnings
```
Luthaf: Same as above

python/metatomic_torchsim/tests/conftest.py
```diff
@@ -6,5 +8,17 @@ def fail_test_with_output(capfd):
     yield
     captured = capfd.readouterr()
     # the code should not print anything to stdout or stderr
+    # except for expected deprecation warnings
```
Luthaf: Same as above

python/metatomic_ase/tests/calculator.py
```diff
@@ -47,11 +47,16 @@ def model():
         energy_unit="eV",
         with_extension=False,
     )
+    # consume the once-per-process quantity deprecation warning from C++
+    captured = capfd.readouterr()
+    if captured.err:
+        assert "ModelOutput.quantity is deprecated" in captured.err
```
Luthaf: We can change the code in the test LJ to remove the need to catch this warning

HaoZeke: https://github.com/metatensor/lj-test/issues/17

Luthaf: Could you do it over there, then we merge and use the new version here directly, instead of needing to update the code twice? Since we always pin lj-test with a specific commit, there should be no downstream issues doing this.

python/metatomic_ase/tests/conftest.py
HaoZeke: I really prefer this, it's much cleaner.

Luthaf: The point of checking warnings explicitly is similar to checking for exceptions. IMO it is not enough to check that "an exception has been thrown at some point", we want to make sure that the exception/warning we expect is produced exactly by the one line of code we expect to throw/emit it

Luthaf (approved):

PR #144, opened 2026-01-26
Add a heat flux wrapper

GardevoirX:
Closes #104 

This is still in a very early stage and subjected to change.

TODO:
- [x] Interface for getting the heat flux from the model
- [x] Checks for the applicability of unfolding the system
- [x] Tests
- [x] Documentations
- [ ] ~Remove Replace the dependency on `vesin.metatomic` with `vesin.torch` and rebuild neighbor list from the un-unfolded system (Hard to build the neighboring information between the replica atoms, e.g. atoms in different image boxes. And `vesin.metatomic` is not supposed to be used in `metatomic`.)~

# Contributor (creator of pull-request) checklist

 - [x] Tests updated (for new features and bugfixes)?
 - [x] Documentation updated (for new features)?
 - [x] Issue referenced (for PRs that solve an issue)?

# Reviewer checklist

 - [ ] CHANGELOG updated with public API or any other important changes?

PR #203, opened 2026-04-17
Create a heat_flux output for each energy variant

Luthaf:
The same model is now able to compute any heat_flux variant, forwarding the energy calculation to the corresponding variant; and even able to compute multiple heat_flux for different variants at the same time.

This is a follow up to #144 

# Reviewer checklist

### DIFF
--- a/python/metatomic_torch/tests/heat_flux.py
+++ b/python/metatomic_torch/tests/heat_flux.py
@@ -8,10 +8,7 @@ from vesin.metatomic import compute_requested_neighbors_from_options
 
 import metatomic_lj_test
 from metatomic.torch import (
-    AtomisticModel,
-    ModelCapabilities,
     ModelEvaluationOptions,
-    ModelMetadata,
     ModelOutput,
     NeighborListOptions,
     systems_to_torch,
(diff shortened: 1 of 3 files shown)


python/metatomic_torch/tests/heat_flux.py
```diff
@@ -123,151 +120,77 @@ def test_heat_flux_wrapper_requested_inputs(model):
...
-        heat_model = torch.jit.script(heat_model)
-
-    results = heat_model([system], evaulation_options, True)[hf_variant].block().values
-    assert torch.allclose(
-        results,
-        torch.tensor(expected, dtype=results.dtype),
-    )
-
+def test_heat_flux_wrapper_calc_heat_flux(model, script, system, variant, expected):
+    heat_flux_model = HeatFlux.wrap(model)
```
GardevoirX, line 135: maybe we can test what happens when we request the two variants at the same time?

GardevoirX (commented):
LGTM, only a small comment

python/metatomic_torch/tests/heat_flux.py
Luthaf: Done!

GardevoirX (approved):
Thanks a lot!

PR #205, opened 2026-04-21
Make sure deprecation warnings don't reach users that can not act on them

Luthaf:
Follow up to #168 

# Reviewer checklist

### DIFF
--- a/metatomic-torch/include/metatomic/torch/model.hpp
+++ b/metatomic-torch/include/metatomic/torch/model.hpp
@@ -43,6 +43,22 @@ public:
         std::string description_
     );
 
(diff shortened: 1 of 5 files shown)


pfebrer:
Who will they reach then? :smile:

Luthaf:
Basically the issue was that end users would get a warnings when loading a model exported with an old version of metatomic. We only want the person exporting the model to get this warning, not someone just trying to run a simulation.

pfebrer:
Ok I would say that this would be the normal way in which things work, end users see the warning, then post an issue and the client code updates to deal with it. I guess in this case since we control all the clients it might not be needed though, so ok 👍

PR #209, opened 2026-04-23
Unify pick_device between C++ and Python

### LINKED ISSUE #176: `pick_device` has different semantics in C++ and Python
The C++ version only returns a DeviceType and no device index since #113, while the Python function allows picking the device index as well since #155. Overall, I think we should handle this more cleanly, it seems wrong to have diverging versions of this function.

PicoCentauri:
Fixes #176

`pick_device` now returns a `torch::Device` (allowing for a user supplied index)                                                                                                                      in both C++ and Python, instead of a bare `DeviceType` in C++ and a reconstructed                                                                                                             string in Python.

The `pick_device_pywrapper` in C++ is removed.

[📚 Download documentation for this pull-request](https://nightly.link/metatensor/metatomic/actions/artifacts/6602854655.zip)

### DIFF
--- a/metatomic-torch/CHANGELOG.md
+++ b/metatomic-torch/CHANGELOG.md
@@ -29,6 +29,10 @@ a changelog](https://keepachangelog.com/en/1.1.0/) format. This project follows
 
 ### Changed
 
+- `pick_device` now returns a `torch::Device` (instead of `torch::DeviceType`)
+  in C++, and a `torch.device` (instead of a `str`) in Python. Any device index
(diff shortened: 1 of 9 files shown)


Luthaf:
I'm a bit concerned by this being a breaking change, and requiring again to scramble and update all engines … Maybe we should keep this open for now and bundle it together with other breaking changes later?

PicoCentauri:
> I'm a bit concerned by this being a breaking change, and requiring again to scramble and update all engines … Maybe we should keep this open for now and bundle it together with other breaking changes later?

Agree! But I think it is cleaner and better now.

PR #210, opened 2026-04-23
Deprecate plural outputs, unify inputs and outputs

Luthaf:
Standard inputs and outputs are now named "standard quantities"; and all names use the singular version of the word (`mass` instead of `masses`)

Most of the code in this PR is handling backward compatibility. For outputs, AtomisticModel will automatically transform between the engine requested outputs and the model capabilties, sending warnings if any of them uses deprecated names.

For inputs, the engine must opt-in using the new names (since we don't want to duplicate inputs in the systems), and AtomisticModel will transform the provided inputs back to what the model wants internally.

# Reviewer checklist


docs/src/engines/ipi.rst
```diff
@@ -15,10 +15,11 @@ i-PI
...
 
-The :ref:`energy <energy-output>`, non-conservative :ref:`forces
-<non-conservative-forces-output>` and :ref:`stress <non-conservative-stress-output>`
-outputs are supported, and can be used to run path integral simulations, incorporating
-integral simulations, incorporating quantum nuclear effects in the statistical sampling.
+The :ref:`energy <energy-quantity>`, non-conservative :ref:`forces
+<non-conservative-force-quantity>` and :ref:`stress
+<non-conservative-stress-quantity>` outputs are supported, and can be used to
+run path integral simulations, incorporating integral simulations, incorporating
+quantum nuclear effects in the statistical sampling.
```
PicoCentauri:
```suggestion
run path integral simulations, incorporating integral simulations and
quantum nuclear effects in the statistical sampling.
```

docs/src/quantities/feature.rst
```diff
@@ -82,10 +84,9 @@ The following simulation engines can use the ``"features"`` output:
...
     |plumed-logo|
 
-Gradients of Features
----------------------
 
-As for the :ref:`energy <energy-output-gradients>`, features are typically used
-with automatic differentiation for the gradients. Explicit gradients could be
-allowed if you have a use case for them, but are currently not implemented until
-they are fully specified.
+Gradients of feature
```
PicoCentauri: This reads a bit weird.

docs/src/quantities/feature.rst
```diff
@@ -82,10 +84,9 @@ The following simulation engines can use the ``"features"`` output:
...
+--------------------
+
+Features are typically used with automatic differentiation for the gradients,
+and explicit gradients are not currently specified.
```
PicoCentauri: Maybe also use singular here:


```suggestion
Gradients of a the feature quantity are typically used with automatic differentiation. Explicit gradients are not currently specified.
```

docs/src/quantities/index.rst
```diff
@@ -0,0 +1,171 @@
...
+metadata when handling the same quantity. If your model returns one of the
+quantity defined in this documentation as output; or use them as input; then it
+must follow the metadata structure described here.
+
+If you need other quantities as inputs or outputs, you should use custom
+quantity with a name containing ``::``, such as ``my_code::my_quantity``. For
+such custom quantity, you are free to use any relevant metadata structure, but
+if multiple people are using the same quantity, they are encouraged to come
+together, define the metadata schema they need and add a new section to these
+pages.
```
PicoCentauri, line 17: Should we mention here that they are usually in singular...

docs/src/quantities/index.rst
```diff
@@ -0,0 +1,171 @@
...
+  non_conservative
+  mass
+  position
+  momentum
+  velocity
+  charge
+  heat_flux
+  spin_multiplicity
+  feature
+  variants
```
PicoCentauri, line 33:
```suggestion
  variant
```

?

docs/src/quantities/index.rst
```diff
@@ -0,0 +1,171 @@
...
+
+Variants
+^^^^^^^^
+
+Models can define variants of any quantity, for example to provide the same
```
PicoCentauri, line 38:
```suggestion
Models can define a variant of any quantity, for example to provide the same
```

docs/src/quantities/index.rst
```diff
@@ -0,0 +1,171 @@
...
+
+      Atomic masses
+
+    .. grid-item-card:: Position
+      :link: position-quantity
+      :link-type: ref
+
+      .. image:: /../static/images/position-quantity.png
+
+      Atomic positions predicted by the model, to be used in ML-driven simulations.
```
PicoCentauri, line 112: To be in line with the nc descriptions.

```suggestion
      Atomic positions directly predicted by the model
```

docs/src/quantities/non_conservative.rst
```diff
@@ -0,0 +1,169 @@
+.. _non-conservative-force-quantity:
+
+Non-conservative force
+^^^^^^^^^^^^^^^^^^^^^^
+
+Non-conservative forces are forces that are not calculated as the negative
```
PicoCentauri, line 6:
```suggestion
Non-conservative forces are not calculated as the negative
```

docs/src/quantities/non_conservative.rst
```diff
@@ -0,0 +1,169 @@
...
+gradient of a potential energy function, but rather directly predicted by the
+model, without going through the potential energy. These are generally faster to
```
PicoCentauri:
```suggestion
model, without going through the potential energy. These forces are generally faster to
```

docs/src/quantities/non_conservative.rst
```diff
@@ -0,0 +1,169 @@
...
+    dynamics engine, metatomic models might predict a non zero total force. You
+    should consider removing this total force to prevent drift in your
+    simulations.
+
+.. _non-conservative-stress-quantity:
+
+Non-conservative stress
+^^^^^^^^^^^^^^^^^^^^^^^
+
+Similar to the forces, the non-conservative stress is a stress tensor that is
```
PicoCentauri:
```suggestion
Similar to the force, the non-conservative stress is a stress tensor that is
```

docs/src/conf.py
```diff
@@ -135,6 +135,7 @@ def setup(app):
     "myst_parser",
     "sphinx_design",
     "chemiscope.sphinx",
+    "sphinx_reredirects",
```
PicoCentauri, line 138: wow nice package

metatomic-torch/src/quantities.cpp
```diff
@@ -744,154 +788,213 @@ void metatomic_torch::check_outputs(
...
+        } else if (base == "position") {
+            check_position(value, systems, request);
+        } else if (base == "momentum") {
+            check_momentum(value, systems, request);
+        } else if (base == "mass") {
+            check_mass(value, systems, request);
+        } else if (base == "velocity") {
+            check_velocity(value, systems, request);
+        } else if (base == "charge") {
+            check_charge(value, systems, request);
```
PicoCentauri, line 829: Do we really need functions for all of them independently? I would say the logic is very similar for most of them, no?

docs/src/quantities/index.rst
Luthaf: I don't think we have to, we can discuss this whenever a PR comes in. If this becomes an issue we can start having docs for how to add new quantities

docs/src/quantities/index.rst
Luthaf: no, there are multiple variants of the same quantity

metatomic-torch/src/quantities.cpp
Luthaf: Yes, but IMO having them as individual functions make it a lot easier to only modify one of them. We already share multiple sub-utility functions to check parts of the TensorMap

PicoCentauri: Okay fine with me.

PicoCentauri (approved):
LGTM

PR #211, opened 2026-04-24
Extend `non_conservative` to a 4-way string mode selector in ASE and TorchSim

PicoCentauri:
Matches the `non_conservative` interface already implemented in the LAMMPS metatomic                                                                                                                  pair style (metatensor/lammps#47). 

The `non_conservative` parameter was a `bool` in both `MetatomicCalculator` and `MetatomicModel`. It is                                                                                                    now a string with four accepted values `"off"` (default), `"on"`, `"forces"` and `"stress"`.

# Reviewer checklist

[📚 Download documentation for this pull-request](https://nightly.link/metatensor/metatomic/actions/artifacts/6620627680.zip)

### DIFF
--- a/python/metatomic_ase/src/metatomic_ase/_calculator.py
+++ b/python/metatomic_ase/src/metatomic_ase/_calculator.py
@@ -2,7 +2,7 @@ import logging
 import os
 import pathlib
 import warnings
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Literal, Optional, Union
 
 import metatensor.torch as mts
 import numpy as np
@@ -125,7 +125,7 @@ class MetatomicCalculator(ase.calculators.calculator.Calculator):
         check_consistency=False,
         device=None,
         variants: Optional[Dict[str, Optional[str]]] = None,
-        non_conservative=False,
+        non_conservative: Union[bool, Literal["forces", "stress"]] = False,
         do_gradients_with_energy=True,
         uncertainty_threshold=0.1,
     ):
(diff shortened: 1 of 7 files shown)


frostedoyster:
Is this what we were talking about yesterday? I don't know if we need it because doing either conservatively (forces or stresses) has the same cost as doing both conservatively

I.e. C forces + NC stresses has the same cost as C forces + C stresses

Of course it's still useful if one wants to test the effects of turning one off, but it's such a niche curiosity that IMO one should hack it locally

Luthaf:
My understanding is more that this is about allowing models that don't offer both outputs (for example because they were not trained with stress)

python/metatomic_ase/src/metatomic_ase/_calculator.py
```diff
@@ -125,7 +125,7 @@ def __init__(
         check_consistency=False,
         device=None,
         variants: Optional[Dict[str, Optional[str]]] = None,
-        non_conservative=False,
+        non_conservative="off",
```
Luthaf: I think I'd prefer keeping backward compatibility by making this 

```suggestion
        non_conservative:Union[bool, Literal["forces", "stress"]]=False,
```

PicoCentauri: I think this is making the API unclear if one allows `"on"` or `True` but to keep backward compatibility I agree we can change this.

PicoCentauri: And following #210 should we call it `"force"` or `"forces"`?

Luthaf: > I think this is making the API unclear if one allows "on" or True

Yes, there should only be `True`, `False`, `"force"`, `"stress"`, no `"on"` here!

> And following https://github.com/metatensor/metatomic/pull/210 should we call it "force" or "forces"?

I'd say "forces" is fine here, this is what you used in LAMMPS, right?

PicoCentauri: Okay I changed the synopsis.

Yes in LAMMPS we have `<on/off/forces/stress>`

Luthaf (approved):
Thanks!

PR #212, opened 2026-04-24
Skip constrained ase.atoms in forward

### LINKED ISSUE #206: Consider `ase.Atoms.constraints` for calculations.
ASE has an atom property called [constraints](https://wiki.fysik.dtu.dk/~askhl/ase-doc/ase/constraints.html) omitting atom updates during optimisations and dynamics. Currently, we compute all forces even if these constraints are present. It is not problematic as the optimizer will do a correct behaviour, but we do unnecessary calculations. I think we should consider `constraints` when doing inference. Since we already have the `selected_atoms`, this should not be too difficult.

Pinging also @prajwal-pisal as he requested this.

PicoCentauri:
Fixes #206

Particles in an `ase.Atoms` object that are part of a `FixAtoms` constraint are now                                                                                                                  
excluded from `selected_atoms` during model inference. avoiding redundant per-atom                                                                                                                     
output computation for atoms whose forces will be zeroed by the optimizer anyway.  

# Reviewer checklist

[📚 Download documentation for this pull-request](https://nightly.link/metatensor/metatomic/actions/artifacts/6628203233.zip)

### DIFF
--- a/python/metatomic_ase/CHANGELOG.md
+++ b/python/metatomic_ase/CHANGELOG.md
@@ -16,6 +16,10 @@ follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased](https://github.com/metatensor/metatomic/)
 
+### Added
+
+- `MetatomicCalculator` is skipping model output computation for spatially fixed atoms.
+
 ## [Version 0.1.0](https://github.com/metatensor/metatomic/releases/tag/metatomic-ase-v0.1.0) - 2026-03-25
 
 - `metatomic-ase` is now a standalone package, containing the ASE integration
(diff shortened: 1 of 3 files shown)


ceriottm:
I'm a bit confused/concerned here. Even if atoms are constrained, their per-atom energy will contribute to the forces on non-constrained atoms, so we can't really drop them from the calculation

PicoCentauri:
We don't remove them from the calculations. Fixed atoms are still neighbors and contribute.

ceriottm:
> We don't remove them from the calculations. Fixed atoms are still neighbors and contribute.

But they don't have their own V_i, that contributes to forces on their neighbors

Luthaf:
True, thanks for the catch @ceriottm! I think we both had a mental model with pair potentials, and did not consider higher body-order 😅 

@PicoCentauri The problem is that the force on atom j `f_j` is `d V / dr_j`, and `V = sum_i V_i`. So in general the force on atom `j` arises from gradients of the energy of all atoms which have `j` as a neighbor, not from the energy on atom `j`.

We could ignore the fixed atoms/skip them in `selected_atoms` when computing forces directly through `non_conservative_force`, but at this point if might become more of a footgun/added complexity in the code for very minimal gains.

PR #219, opened 2026-05-04
DFT-D3 wrapper

GardevoirX:
This PR aims to introduce a DFT-D3 wrapper, which can be wrapped outside of MLIPs in `AtomisticModel`. The wrapper currently only supports D3(BJ).

This wrapper calculates the energy correction of the input structure, and add it to the energy calculated by the wrapped MLIP. In this way, and through the auto-differentiation, the D3-corrected forces and stresses (if with full-PBC) can be automatically calculated.

I have tested this wrapper on MAD dataset. The energies, forces, and stresses are in good agreement with another wrapper implemented with the DFT-D3 provided in the nvalchemiops package.

The overhead is about 15%-35% for s models on systems ranging from 600 atoms to 4k atoms. Mainly from the neighbor list calculation, the actual overheads of forward and backward are only ~3% and ~8%.

Remaining issues:
- [x] The D3 parameters shared across different functionals is now downloaded (through [this function](https://github.com/NVIDIA/nvalchemi-toolkit-ops/blob/619af5b0b5b19f1f78bd05fb1e3ac9143fc4b631/examples/dispersion/utils.py#L553)) and passed to the wrapper by the user. Maybe we need to incorporate this into the wrapper
- [x] The D3 parameters are in atomic units, so maybe a unit conversion helper is needed
- [x] Maybe also the correction to the non-conservative forces and stresses
- [x] Sometimes we might want to put correction to only a part of the atoms in the system (requested by @cesaremalosso, Ref: https://pubs.acs.org/doi/full/10.1021/acs.jpclett.3c00856 )
- [x] Should we update the interaction range in `ModelCapability`? Currently it's updated
- [ ] Should we force users to use the same cutoff as the DFT-D3 calculation for calculating the coordination numbers of atoms?
- [ ] Correctly cite the packages used during the development (tad-dftd3, nvalchemiops) (Maybe)

# Reviewer checklist


python/metatomic_ase/src/metatomic_ase/_neighbors.py
```diff
@@ -134,6 +134,7 @@ def _compute_requested_neighbors_nvalchemi(systems, requested_options):
                 options.engine_cutoff("angstrom"),
                 cell=system.cell,
                 pbc=system.pbc,
+                max_neighbors=16384,
```
Luthaf: Why is this required? and why this value?

python/metatomic_torch/tests/dftd3.py
```diff
@@ -0,0 +1,390 @@
...
+SIGMA = 3.405
+EPSILON = 0.01032
+
+# 7-point reference grid matching the standard Grimme tables.
+_REF_GRID = 7
+
+_D3_REFERENCE = {
+    "default": {
+        "energy": -1.0188759583539984,
+        "forces": np.fromstring(
```
Luthaf: why fromstring?? You can put the values as Python code

python/metatomic_torch/tests/dftd3.py
```diff
@@ -0,0 +1,390 @@
...
+        sigma=SIGMA,
+        epsilon=EPSILON,
+        length_unit="Angstrom",
+        energy_unit="eV",
+        with_extension=True,
+    )
+
+
+@pytest.fixture
+def model_pure():
```
Luthaf: Instead of using LJ, we could test D3 on top of a model that returns 0 energy and non-conservative forces

python/metatomic_torch/tests/dftd3.py
```diff
@@ -0,0 +1,390 @@
...
+        )
+    )
+    return block.values.squeeze(-1).detach().cpu().numpy()[order]
+
+
+def test_dftd3_default_cutoffs_use_grimme(model_with_extension):
+    """When the caller does not specify cutoffs, the wrapper must default to
+    the Grimme values (50 / 25 Bohr) converted into the model's length unit
+    (here Angstrom)."""
+    wrapper = DFTD3(
```
Luthaf: like for the heat flux, this use `DFTD3.wrap()` everywhere (since it is the intended API for the end user)

python/metatomic_torch/metatomic/torch/dftd3.py
```diff
@@ -0,0 +1,597 @@
...
+    ``"energy/pbe"``) listed in ``damping_params``, the wrapper adds the D3
+    energy as a differentiable tensor: ``E_corrected = E_base + E_D3``.
+
+    The D3 energy is implemented in pure PyTorch and is naturally
+    differentiable: ``torch.autograd`` flows from the corrected energy back to
+    ``positions`` and ``cell`` through the neighbor-list distances (vesin
+    registers an autograd hook so the neighbor vectors carry gradients), so
+    the standard ``-∇E`` optimization path
+    (``MetatomicCalculator(..., non_conservative=False)``) yields conservative
+    forces and stress that include the D3 contribution.
```
Luthaf:
```suggestion
    ``positions`` and ``cell`` through the neighbor list distances.
```

python/metatomic_torch/metatomic/torch/dftd3.py
```diff
@@ -0,0 +1,597 @@
...
+
+    The wrapper does **not** modify ``non_conservative_forces[/<variant>]``
+    or ``non_conservative_stress[/<variant>]`` outputs the base model may
+    expose. Those remain the base model's direct-predict outputs (no D3
+    contribution). Use the autograd path for D3-corrected forces / stress.
```
Luthaf: I think it would be nice to also correct these if we can

python/metatomic_torch/metatomic/torch/dftd3.py
```diff
@@ -0,0 +1,597 @@
...
+            cn_cutoff = _D3_CN_CUTOFF_BOHR * bohr_to_model
+
+        if cutoff <= 0.0:
+            raise ValueError("DFTD3 cutoff must be positive, got " + str(cutoff))
+        if cn_cutoff <= 0.0:
+            raise ValueError("DFTD3 cn_cutoff must be positive, got " + str(cn_cutoff))
+
+        max_atomic_type = max(capabilities.atomic_types)
+        if max_atomic_type >= rcov.shape[0]:
+            warn(
```
Luthaf:
```suggestion
            warnings.warn(
```

python/metatomic_torch/metatomic/torch/dftd3.py
```diff
@@ -0,0 +1,597 @@
...
+        if c6.ndim != 4:
+            raise ValueError("'c6' must be 4D, got shape " + str(tuple(c6.shape)))
+        if cn_ref.ndim != 2:
+            raise ValueError(
+                "'cn_ref' must be 2D, got shape " + str(tuple(cn_ref.shape))
+            )
+        if c6.shape[0] != c6.shape[1]:
+            raise ValueError(
+                "'c6' must be square in its first two axes, got shape "
+                + str(tuple(c6.shape))
```
Luthaf: Can you use f-strings instead of string concatenation for all errors?

python/metatomic_torch/metatomic/torch/dftd3.py
```diff
@@ -0,0 +1,597 @@
...
+            damping_params=damping_params,
+            cutoff=cutoff,
+            cn_cutoff=cn_cutoff,
+        )
+
+        capabilities = model.capabilities()
+        supported_devices = [
+            device
+            for device in capabilities.supported_devices
+            if device in ["cpu", "cuda"]
```
Luthaf: Why this filter? if the code is pure torch, any device should be supported

python/metatomic_torch/metatomic/torch/dftd3.py
```diff
@@ -0,0 +1,597 @@
...
+            if outputs[energy_key].sample_kind != "system":
+                raise NotImplementedError(
+                    "DFTD3 does not support per-atom corrected energies"
+                )
+            need_variants.append(energy_key)
+
+        if selected_atoms is not None and len(need_variants) > 0:
+            raise NotImplementedError(
+                "DFTD3 does not support 'selected_atoms' for corrected outputs"
+            )
```
Luthaf: This will prevent using this code with LAMMPS, so I think it would be worth implementing =)

Luthaf:
> The D3 parameters shared across different functionals is now downloaded (through [this function](https://github.com/NVIDIA/nvalchemi-toolkit-ops/blob/619af5b0b5b19f1f78bd05fb1e3ac9143fc4b631/examples/dispersion/utils.py#L553)) and passed to the wrapper by the user. Maybe we need to incorporate this into the wrapper

Yes, ideally running this code should not require internet access

> The D3 parameters are in atomic units, so maybe a unit conversion helper is needed

Isn't this implemented? 

> Maybe also the correction to the non-conservative forces and stresses

Yes!

> Sometimes we might want to put correction to only a part of the atoms in the system (requested by @cesaremalosso )

This should be doable, but how would it look when working with heterogeneous systems?

> Should we update the interaction range in ModelCapability? Currently it's updated

Yes!

> Should we force users to use the same cutoff as the DFT-D3 calculation for calculating the coordination numbers of atoms?

Why would someone want to do or not do this? I don't quite understand the implications.

python/metatomic_ase/src/metatomic_ase/_neighbors.py
GardevoirX: When testing on mad dataset, there were some errors related to the number of neighbors. Chatty lifted this value twice to get everything passed. It chose this value because it satisfies 2^n, I think. If we are interested in which structure led to the error, I can do more tests

Luthaf: Yes please! I think we could set this to a similar value as the one we use in vesin, which is based on the cutoff + number of atoms.

python/metatomic_torch/tests/dftd3.py
```diff
@@ -1,109 +1,297 @@
...
+        ModelMetadata(),
+        ModelCapabilities(
+            outputs={
+                "energy": ModelOutput(
+                    sample_kind="system", unit="eV", description="D3-corrected energy"
+                ),
+                "energy/doubled": ModelOutput(
+                    sample_kind="system",
+                    unit="eV",
+                    description="D3-corrected energy for doubled head",
```
Luthaf: These are not yet D3 corrected

python/metatomic_torch/tests/dftd3.py
```diff
@@ -146,41 +334,135 @@ def _damping(**overrides):
     return p
 
 
-def _eval(model, atoms, outputs, check_consistency=True):
-    calc = MetatomicCalculator(
+def _damping_with_direct():
```
Luthaf: this is maybe not worth a function …

python/metatomic_torch/metatomic/torch/dftd3.py
```diff
@@ -22,6 +23,35 @@
...
+    :class:`DFTD3`: ``rcov`` and ``r4r2`` in Bohr, ``c6`` in
+    Hartree * Bohr^6, and dimensionless ``cn_ref`` values. The table layout is
+    the current pure-PyTorch wrapper layout: ``c6`` has shape ``(Z, Z, M, M)``
+    and ``cn_ref`` has shape ``(Z, M)``.
+    """
+    try:
+        import numpy as np
+    except ImportError as e:
+        raise RuntimeError("loading packaged DFT-D3 parameters requires numpy") from e
+
```
Luthaf: We have a numpy dependency through metatensor-core, so we can assume it is here (or just add it to this project deps as well).

python/metatomic_torch/metatomic/torch/dftd3.py
```diff
@@ -30,29 +60,42 @@ class DFTD3(torch.nn.Module):
     dispersion correction to its energy output(s). The three-body correction
     term is **not** included.
 
-    The wrapper can correct multiple energy variants at once, each with its
+    The wrapper can correct multiple output variants at once, each with its
```
Luthaf, line 60:
```suggestion
    The wrapper can correct multiple outputs (including multiple variants of a given output) at once, each with its
```

python/metatomic_torch/metatomic/torch/dftd3.py
```diff
@@ -225,6 +305,28 @@ def __init__(
     def requested_neighbor_lists(self) -> List[NeighborListOptions]:
         return self._requested_neighbor_lists + [self._neighbor_list]
 
+    @staticmethod
+    def _register_direct_output(
```
Luthaf:
```suggestion
    def _register_non_conservative_output(
```

python/metatomic_torch/tests/dftd3.py
```diff
@@ -1,109 +1,297 @@
...
 ATOMIC_NUMBER = 18
-LJ_CUTOFF = 7.0
 D3_CUTOFF = 5.0
-SIGMA = 3.405
-EPSILON = 0.01032
 
 # 7-point reference grid matching the standard Grimme tables.
 _REF_GRID = 7
 
 _D3_REFERENCE = {
```
Luthaf, line 35: Please add some explanation about this, what is contained in the arrays, how it was computed, etc.

Luthaf (commented):
This will need more review, just sending the notes I took during our discussion

GardevoirX:
Here I upload the parity plot of the current version of d3 wrapper against `simple-dftd3`
image
and the script
```python
#!/usr/bin/env python3
from __future__ import annotations

import argparse
import os
import time
from concurrent.futures import ThreadPoolExecutor
from pathlib import Path

import ase.io
import numpy as np
import torch

from metatomic.torch import load_atomistic_model
from metatomic.torch.dftd3 import DFTD3
from metatomic_ase import MetatomicCalculator


ROOT = Path('/home/qxu/repos/metatomic')
DAMPING = {'s6': 1.0, 's8': 0.7875, 'a1': 0.4289, 'a2': 4.4407}
CUTOFF = 15.0
CN_CUTOFF = 15.0


def _stress_allowed(atoms) -> bool:
    return bool(np.all(atoms.pbc)) and abs(atoms.cell.volume) > 0.0


def _batched(frames, batch_size: int, max_atoms: int):
    batch = []
    atoms_in_batch = 0
    stress_allowed = False

    for frame in frames:
        frame_stress_allowed = _stress_allowed(frame[1])
        n_atoms = len(frame[1])

        if batch and frame_stress_allowed != stress_allowed:
            yield batch
            batch = []
            atoms_in_batch = 0
        if batch and max_atoms > 0 and atoms_in_batch + n_atoms > max_atoms:
            yield batch
            batch = []
            atoms_in_batch = 0

        batch.append(frame)
        atoms_in_batch += n_atoms
        stress_allowed = frame_stress_allowed

        if len(batch) == batch_size:
            yield batch
            batch = []
            atoms_in_batch = 0

    if batch:
        yield batch


def _simple_dftd3(atoms):
    from dftd3.ase import DFTD3 as SimpleDFTD3

    atoms = atoms.copy()
    atoms.calc = SimpleDFTD3(
        damping='d3bj',
        params_tweaks={**DAMPING, 's9': 0.0},
        realspace_cutoff={'disp2': CUTOFF, 'disp3': CUTOFF, 'cn': CN_CUTOFF},
    )

    energy = atoms.get_potential_energy()
    forces = atoms.get_forces()
    stress = atoms.get_stress(voigt=False) if _stress_allowed(atoms) else None
    return energy, forces, stress


def _metrics(reference: np.ndarray, value: np.ndarray) -> str:
    delta = np.asarray(value, dtype=float).reshape(-1) - np.asarray(
        reference, dtype=float
    ).reshape(-1)
    if delta.size == 0:
        return 'not available'
    return (
        f'RMSE={np.sqrt(np.mean(delta * delta)):.3e}  '
        f'MAE={np.mean(np.abs(delta)):.3e}  '
        f'max={np.max(np.abs(delta)):.3e}'
    )


def _plot_panel(ax, reference, value, title: str, max_points: int):
    reference = np.asarray(reference, dtype=float).reshape(-1)
    value = np.asarray(value, dtype=float).reshape(-1)
    finite = np.isfinite(reference) & np.isfinite(value)
    reference = reference[finite]
    value = value[finite]

    if reference.size == 0:
        ax.set_title(title)
        ax.text(0.5, 0.5, 'not available', ha='center', va='center')
        ax.set_axis_off()
        return

    if reference.size > max_points:
        rng = np.random.default_rng(0)
        keep = rng.choice(reference.size, size=max_points, replace=False)
        reference_plot = reference[keep]
        value_plot = value[keep]
    else:
        reference_plot = reference
        value_plot = value

    lo = min(reference.min(), value.min())
    hi = max(reference.max(), value.max())
    pad = 0.03 * (hi - lo if hi > lo else max(abs(hi), 1.0))
    lo -= pad
    hi += pad

    ax.scatter(reference_plot, value_plot, s=6, alpha=0.35, linewidths=0)
    ax.plot([lo, hi], [lo, hi], color='black', linewidth=1)
    ax.set_xlim(lo, hi)
    ax.set_ylim(lo, hi)
    ax.set_title(title)
    ax.set_xlabel('simple-dftd3')
    ax.set_ylabel('metatomic wrapper correction')
    ax.text(
        0.04,
        0.96,
        f'N={reference.size}\n{_metrics(reference, value)}',
        transform=ax.transAxes,
        va='top',
        ha='left',
        fontsize=9,
        bbox={'facecolor': 'white', 'alpha': 0.75, 'edgecolor': 'none'},
    )


def _plot(path: Path, ref_energy, value_energy, ref_forces, value_forces, ref_stress, value_stress, max_points: int):
    import matplotlib

    matplotlib.use('Agg')
    import matplotlib.pyplot as plt

    fig, axes = plt.subplots(1, 3, figsize=(14, 4.2), constrained_layout=True)
    _plot_panel(axes[0], ref_energy, value_energy, 'Energy (eV)', max_points)
    _plot_panel(axes[1], ref_forces, value_forces, 'Force components (eV/A)', max_points)
    _plot_panel(axes[2], ref_stress, value_stress, 'Stress components (eV/A^3)', max_points)
    fig.savefig(path, dpi=220)
    plt.close(fig)


def main() -> int:
    parser = argparse.ArgumentParser()
    parser.add_argument('--xyz', default=str(ROOT / 'mad-train.xyz'))
    parser.add_argument('--model', default=str(ROOT / 'pet-mad-s-v1.0.2.pt'))
    parser.add_argument('--output', default='dftd3-simple-parity.png')
    parser.add_argument('--npz', default='dftd3-simple-parity.npz')
    parser.add_argument('--device', choices=['cpu', 'cuda'], default='cuda')
    parser.add_argument('--batch-size', type=int, default=200)
    parser.add_argument('--max-atoms-per-batch', type=int, default=1800)
    parser.add_argument('--reference-workers', type=int, default=8)
    parser.add_argument('--max-points', type=int, default=200_000)
    args = parser.parse_args()

    os.environ['OMP_NUM_THREADS'] = '1'
    os.environ['OPENBLAS_NUM_THREADS'] = '1'
    os.environ['MKL_NUM_THREADS'] = '1'

    model = load_atomistic_model(args.model)
    wrapped = DFTD3.wrap(
        model,
        damping_params={'energy': DAMPING},
        cutoff=CUTOFF,
        cn_cutoff=CN_CUTOFF,
    )
    base_calc = MetatomicCalculator(
        model,
        device=args.device,
        check_consistency=False,
        non_conservative=False,
        uncertainty_threshold=None,
    )
    wrapped_calc = MetatomicCalculator(
        wrapped,
        device=args.device,
        check_consistency=False,
        non_conservative=False,
        uncertainty_threshold=None,
    )

    frame_ids = []
    stress_frame_ids = []
    reference_energy = []
    reference_forces = []
    reference_stress = []
    wrapper_energy = []
    wrapper_forces = []
    wrapper_stress = []
    reference_time = 0.0
    wrapper_time = 0.0

    frames = enumerate(ase.io.iread(args.xyz, ':'))
    batches = _batched(frames, args.batch_size, args.max_atoms_per_batch)

    for batch in batches:
        indices = [index for index, _ in batch]
        atoms_batch = [atoms for _, atoms in batch]
        frame_ids.extend(indices)
        batch_has_stress = all(_stress_allowed(atoms) for atoms in atoms_batch)

        if args.device == 'cuda':
            torch.cuda.synchronize()
        start = time.perf_counter()
        base = base_calc.compute_energy(atoms_batch, compute_forces_and_stresses=True)
        wrapped_out = wrapped_calc.compute_energy(
            atoms_batch, compute_forces_and_stresses=True
        )
        if args.device == 'cuda':
            torch.cuda.synchronize()
        wrapper_time += time.perf_counter() - start

        wrapper_energy.append(
            np.asarray(wrapped_out['energy'], dtype=float).reshape(-1)
            - np.asarray(base['energy'], dtype=float).reshape(-1)
        )
        wrapper_forces.append(
            np.concatenate(wrapped_out['forces'], axis=0)
            - np.concatenate(base['forces'], axis=0)
        )
        if batch_has_stress:
            wrapper_stress.append(
                np.asarray(wrapped_out['stress'], dtype=float)
                - np.asarray(base['stress'], dtype=float)
            )

        start = time.perf_counter()
        with ThreadPoolExecutor(max_workers=args.reference_workers) as pool:
            references = list(pool.map(_simple_dftd3, atoms_batch))
        reference_time += time.perf_counter() - start

        for index, (energy, forces, stress) in zip(indices, references, strict=True):
            reference_energy.append(energy)
            reference_forces.append(forces)
            if stress is not None:
                stress_frame_ids.append(index)
                reference_stress.append(stress.reshape(1, 3, 3))

    reference_energy = np.asarray(reference_energy)
    reference_forces = np.concatenate(reference_forces, axis=0)
    reference_stress = (
        np.concatenate(reference_stress, axis=0)
        if reference_stress
        else np.empty((0, 3, 3))
    )
    wrapper_energy = np.concatenate(wrapper_energy, axis=0)
    wrapper_forces = np.concatenate(wrapper_forces, axis=0)
    wrapper_stress = (
        np.concatenate(wrapper_stress, axis=0)
        if wrapper_stress
        else np.empty((0, 3, 3))
    )

    print(f'frames={len(frame_ids)}')
    print(f'stress_frames={len(stress_frame_ids)}')
    print(f'simple_dftd3_time_s={reference_time:.3f}')
    print(f'wrapper_minus_base_time_s={wrapper_time:.3f}')
    print('energy:', _metrics(reference_energy, wrapper_energy))
    print('forces:', _metrics(reference_forces, wrapper_forces))
    print('stress:', _metrics(reference_stress, wrapper_stress))

    _plot(
        Path(args.output),
        reference_energy,
        wrapper_energy,
        reference_forces,
        wrapper_forces,
        reference_stress,
        wrapper_stress,
        args.max_points,
    )
    np.savez_compressed(
        args.npz,
        frame_ids=np.asarray(frame_ids, dtype=np.int64),
        stress_frame_ids=np.asarray(stress_frame_ids, dtype=np.int64),
        reference_energy=reference_energy,
        wrapper_energy=wrapper_energy,
        reference_forces=reference_forces,
        wrapper_forces=wrapper_forces,
        reference_stress=reference_stress,
        wrapper_stress=wrapper_stress,
        simple_dftd3_time_s=np.asarray(reference_time),
        wrapper_minus_base_time_s=np.asarray(wrapper_time),
    )
    print(f'plot: {args.output}')
    print(f'data: {args.npz}')
    return 0


if __name__ == '__main__':
    raise SystemExit(main())
```
To run this script, one should compile simple-dftd3 from source, because one would have to modify this line
https://github.com/dftd3/simple-dftd3/blob/6f0b06fbfa8653a23ca55c453772ce3af4420706/src/dftd3/cutoff.f90#L127
from `spread(periodic, 2, 3)` to `spread(periodic, 1, 3)` which, imho, is the correct implementation

python/metatomic_torch/setup.py
```diff
@@ -350,6 +350,7 @@ def create_version_number(version):
                 "metatomic/torch/torch-*/bin/*",
                 "metatomic/torch/torch-*/lib/*",
                 "metatomic/torch/torch-*/include/*",
-            ]
+            ],
+            "metatomic.torch": ["data/*.npz"],
```
Luthaf: this seems wrong, should be 
```suggestion
            "metatomic-torch": ["metatomic/torch/data/*"],
```

python/metatomic_torch/tests/dftd3.py
```diff
@@ -0,0 +1,842 @@
...
+                        torch.tensor([[0]], dtype=torch.int64, device=device),
+                    ),
+                )
+                blocks = torch.jit.annotate(List[TensorBlock], [block])
+                results[name] = TensorMap(keys, blocks)
+
+        return results
+
+    def requested_neighbor_lists(self) -> List[NeighborListOptions]:
+        return torch.jit.annotate(List[NeighborListOptions], [])
```
Luthaf: You don't have to define this function if the model does not request any NL

python/metatomic_torch/tests/dftd3.py
```diff
@@ -0,0 +1,842 @@
...
+    return system
+
+
+def _d3_params():
+    """Synthetic D3 reference tables matching the values used to generate the
+    ground-truth snapshot.
+
+    The new (Z, M) / (Z, Z, M, M) layout collapses to the same effective
+    physics as the old (Z, Z, 5, 5) test fixture: a single Ar-Ar C6 of 100,
+    one valid CN reference at 1.0, and rcov / r4r2 of 1.5 / 2.0.
```
Luthaf: This makes no sense to a new reader of the code

python/metatomic_torch/tests/dftd3.py
```diff
@@ -0,0 +1,842 @@
...
+def test_dftd3_default_cutoffs_use_grimme(model_with_extension):
+    """When the caller does not specify cutoffs, the wrapper must default to
+    the Grimme values (50 / 25 Bohr) converted into the model's length unit
+    (here Angstrom)."""
+    wrapper = DFTD3.wrap(
+        model_with_extension,
+        d3_params=_d3_params(),
+        damping_params={"energy": {"a1": 0.4, "a2": 4.0, "s8": 1.0}},
+    )
+    BOHR_IN_ANGSTROM = 0.5291772105448199
```
Luthaf: Please use metatomic unit conversion to check this

python/metatomic_torch/tests/dftd3.py
```diff
@@ -0,0 +1,842 @@
...
+    wrapper = DFTD3.wrap(
+        model_with_extension,
+        damping_params={"energy": {"a1": 0.4, "a2": 4.0, "s8": 1.0}},
+        cutoff=D3_CUTOFF,
+        cn_cutoff=D3_CUTOFF,
+    )
+
+    nls = wrapper.requested_neighbor_lists()
+    assert len(nls) == 1
+    assert nls[0].requestors() == ["DFTD3"]
```
Luthaf: This test name does not match what is being tested

python/metatomic_torch/tests/dftd3.py
```diff
@@ -0,0 +1,842 @@
...
+
+
+def test_dftd3_selected_atoms_partition_energy(atoms, model_with_extension):
```
Luthaf:
```suggestion
def test_dftd3_selected_atoms(atoms, model_with_extension):
```

python/metatomic_torch/tests/dftd3.py
```diff
@@ -0,0 +1,842 @@
...
+        ),
+    },
+    "shifted": {
+        "energy": -6.195052365042283,
+    },
+}
+
+
+@pytest.fixture
+def model_with_extension():
```
Luthaf:
```suggestion
def model():
```

There are no extensions here

python/metatomic_torch/tests/dftd3.py
```diff
@@ -0,0 +1,842 @@
...
+        atoms,
+        {"energy": ModelOutput(sample_kind="system")},
+    )
+    corrected_energy = float(results["energy"].block().values.item())
+    np.testing.assert_allclose(
+        corrected_energy, _D3_REFERENCE["default"]["energy"], rtol=1e-10, atol=1e-12
+    )
+
+    with pytest.raises(Exception, match="per atom, only globally"):
+        _eval(wrapped, atoms, {"energy": ModelOutput(sample_kind="atom")})
```
Luthaf: This is already tested above
```suggestion
```

python/metatomic_torch/tests/dftd3.py
```diff
@@ -0,0 +1,842 @@
...
+    results, _, _ = _eval(
+        reloaded,
+        atoms,
+        {"energy": ModelOutput(sample_kind="system")},
+    )
+    reloaded_energy = float(results["energy"].block().values.item())
+    assert np.isclose(reloaded_energy, original_energy, atol=1e-5, rtol=1e-5)
+
+
+def test_dftd3_non_conservative_save_and_reload(tmp_path, model_with_extension, atoms):
```
Luthaf: could this test be merged with the one above?

python/metatomic_torch/tests/dftd3.py
```diff
@@ -0,0 +1,842 @@
...
+        rtol=1e-8,
+    )
+
+
+@pytest.mark.parametrize(
+    ("damping_params", "message"),
+    [
+        (
+            {"energy/does_not_exist": {"a1": 0.4, "a2": 4.0, "s8": 1.0}},
+            "wrapped model does not expose",
```
Luthaf: Please match on the full error message. IMO this would be easier to do by explicitly calling the two failing cases one after the other in the function instead of using `@pytest.mark.parametrize`.

python/metatomic_torch/tests/dftd3.py
```diff
@@ -0,0 +1,842 @@
...
+    )
+    np.testing.assert_allclose(
+        wrapped_stress,
+        _D3_REFERENCE["default"]["stress"],
+        atol=1e-12,
+        rtol=1e-8,
+    )
+
+
+def test_dftd3_selected_atoms_non_conservative_outputs(atoms, model_with_extension):
```
Luthaf: Please merge this with the other selected_atoms test

python/metatomic_torch/metatomic/torch/dftd3.py
```diff
@@ -0,0 +1,985 @@
...
+
+# Standard Grimme D3 cutoffs from ``tad_dftd3.defaults`` (atomic units / Bohr).
+# Used as defaults when the caller does not specify cutoffs explicitly; we
+# convert from Bohr to the wrapped model's length unit at construction.
+_D3_DISP_CUTOFF_BOHR = 50.0
+_D3_CN_CUTOFF_BOHR = 25.0
+_D3_PARAMETERS_NPZ = "dftd3_parameters.npz"
+
+
+def load_dftd3_parameters(
```
Luthaf: Since the module is public (no `_` at the start of the name), please make all functions that are not intended for public use private
```suggestion
def _load_dftd3_parameters(
```

python/metatomic_torch/metatomic/torch/dftd3.py
```diff
@@ -0,0 +1,985 @@
...
+    The wrapper can correct multiple output variants at once, each with its
+    own damping parameters. For every energy output key (e.g. ``"energy"`` or
+    ``"energy/pbe"``) listed in ``damping_params``, the wrapper adds the D3
+    energy as a differentiable tensor: ``E_corrected = E_base + E_D3``.
+
+    The D3 energy is implemented in pure PyTorch and is naturally
+    differentiable: ``torch.autograd`` flows from the corrected energy back to
+    ``positions`` and ``cell`` through the neighbor list distances.
+
+    ``damping_params`` can also contain direct output keys such as
```
Luthaf:
```suggestion
    ``damping_params`` can also be specified for non-conservative outputs such as
```

python/metatomic_torch/metatomic/torch/dftd3.py
```diff
@@ -0,0 +1,985 @@
...
+    ``"non_conservative_force/<variant>"`` and
+    ``"non_conservative_stress/<variant>"``. Direct force/stress outputs are
```
Luthaf:
```suggestion
    ``"non_conservative_stress/<variant>"``. Non-conservative force/stress outputs are
```

python/metatomic_torch/metatomic/torch/dftd3.py
```diff
@@ -0,0 +1,985 @@
...
+    corrected only when their output keys are explicitly listed in
+    ``damping_params``; energy damping parameters are not inferred for these
+    outputs. The D3 contribution to direct force/stress outputs is obtained by
+    a local autograd derivative of the same D3 energy expression with respect
+    to the neighbor-vector values.
```
Luthaf: Would it make sense to have an explicit formula for the force instead? Might be faster than relying on autograd

python/metatomic_torch/metatomic/torch/dftd3.py
```diff
@@ -0,0 +1,985 @@
...
+        The D3 correction to ``non_conservative_force[/<variant>]`` and
+        ``non_conservative_stress[/<variant>]`` requires gradient tracking
+        inside ``forward``. These direct corrected outputs can not be computed
+        under ``torch.no_grad()`` or ``torch.inference_mode()``.
+
+    The D3 reference tables (``d3_params``) are shared across variants,
+    matching the convention that the Grimme reference data is functional
+    independent. Damping parameters (``a1``, ``a2``, ``s8``, ...) are
+    provided per variant. All D3 tables and damping parameters **must** be
+    passed in **atomic units**. The wrapper converts the final D3 energy into the
```
Luthaf: Please specify what atomic units means (I guess Bohr+Hartree?)

python/metatomic_torch/metatomic/torch/dftd3.py
```diff
@@ -0,0 +1,985 @@
...
+    _stress_keys: List[str]
+    _force_damping_keys: Dict[str, str]
+    _stress_damping_keys: Dict[str, str]
+    _force_units: Dict[str, str]
+    _stress_units: Dict[str, str]
+
+    def __init__(
+        self,
+        model: AtomisticModel,
+        damping_params: Dict[str, Dict[str, float]],
```
Luthaf, line 118:
```suggestion
        *,
        damping_params: Dict[str, Dict[str, float]],
```

To force all other parameters to be given as keyword args

python/metatomic_torch/metatomic/torch/dftd3.py
```diff
@@ -0,0 +1,985 @@
...
+        :param damping_params: a mapping from an output key to a mapping of damping
+            parameters for that output. Keys can be ``"energy[/<variant>]"``,
+            ``"non_conservative_force[/<variant>]"`` or
+            ``"non_conservative_stress[/<variant>]"``. Each damping map must provide
+            ``a1``, ``a2`` and ``s8``; ``s6`` is optional (defaults to 1.0).
+        :param d3_params: shared DFT-D3 reference tables with keys ``"rcov"``
+            (shape ``(Z,)``), ``"r4r2"`` (shape ``(Z,)``), ``"c6"`` (shape
+            ``(Z, Z, M, M)`` — typically ``M = 5``) and ``"cn_ref"`` (shape
+            ``(Z, M)`` — per-element CN reference grid, with ``-1`` marking
+            absent slots). Tables must be in D3 atomic units. If ``None``,
```
Luthaf: What's `Z`? The max atomic number supported?

python/metatomic_torch/metatomic/torch/dftd3.py
```diff
@@ -0,0 +1,985 @@
...
+    corrected only when their output keys are explicitly listed in
+    ``damping_params``; energy damping parameters are not inferred for these
+    outputs. The D3 contribution to direct force/stress outputs is obtained by
+    a local autograd derivative of the same D3 energy expression with respect
+    to the neighbor-vector values.
+
+    ``selected_atoms`` is supported with the usual domain-decomposition
+    convention: the D3 environment is computed with all atoms in each
+    :class:`System`, while pair energies are split equally between the two pair
+    endpoints and only the shares belonging to selected atoms are added.
```
Luthaf, line 81: Doesn't this mean that this code could support sample_kind = atom?

python/metatomic_torch/metatomic/torch/dftd3.py
```diff
@@ -0,0 +1,985 @@
...
+        damping_method = "bj"
+        for output_key, params in damping_params.items():
+            is_energy = output_key == "energy" or output_key.startswith("energy/")
+            is_force = output_key == "non_conservative_force" or output_key.startswith(
+                "non_conservative_force/"
+            )
+            is_stress = (
+                output_key == "non_conservative_stress"
+                or output_key.startswith("non_conservative_stress/")
+            )
```
Luthaf:
```suggestion
            is_energy = output_key.split("/")[0] == "energy"
            is_force = output_key.split("/")[0] == "non_conservative_force"
            is_stress = output_key.split("/")[0] == "non_conservative_stress"
```

python/metatomic_torch/metatomic/torch/dftd3.py
```diff
@@ -0,0 +1,985 @@
...
+        self._cn_cutoff = cn_cutoff
+        self._neighbor_cutoff = max(cutoff, cn_cutoff)
+
+        self._requested_neighbor_lists = model.requested_neighbor_lists()
+        self._neighbor_list = NeighborListOptions(
+            cutoff=self._neighbor_cutoff,
+            full_list=False,
+            strict=True,
+            requestor="DFTD3",
+        )
```
Luthaf, line 318: Shouldn't the code also request a NL for CN?

python/metatomic_torch/metatomic/torch/dftd3.py
```diff
@@ -0,0 +1,985 @@
...
+        if rcov.shape[0] < c6.shape[0] or r4r2.shape[0] < c6.shape[0]:
+            raise ValueError(
+                f"'rcov' and 'r4r2' must cover at least 'c6' first axis "
+                f"length ({c6.shape[0]}), got {rcov.shape[0]} and {r4r2.shape[0]}"
+            )
+
+    @staticmethod
+    def wrap(
+        model: AtomisticModel,
+        damping_params: Dict[str, Dict[str, float]],
```
Luthaf, line 397:
```suggestion
        *,
        damping_params: Dict[str, Dict[str, float]],
```

python/metatomic_torch/metatomic/torch/dftd3.py
```diff
@@ -0,0 +1,985 @@
...
+        )
+        return offset_tensor[sample_values[:, 0]] + sample_values[:, 1]
+
+    @staticmethod
+    def _energy_block_as_system(block: TensorBlock, n_systems: int) -> TensorBlock:
+        """
+        Called when the energy block is per-atom; sum back to per-system values so we
+        can add the D3 correction.
+        """
+        if block.samples.names == ["system"]:
```
Luthaf: Please use `mts.sum_over_samples`

python/metatomic_torch/metatomic/torch/dftd3.py
GardevoirX: No it just requests a NL for the largest cutoff, and screens the results to get the NL of the smaller cutoff

python/metatomic_torch/metatomic/torch/dftd3.py
GardevoirX: Because the split of the pair energy is arbitrary, I prefer to disable the calculation of the per-atom energies. I just want to enable the usage in lammps by implementing this

python/metatomic_torch/metatomic/torch/dftd3.py
GardevoirX: yes combinations of Bohr and Hartree, but different parameters (`a1`, `a2`, `s6`, `s8`, `rcov` , `r4r2`, `c6`, and `cn_ref`) have different units, should we specify all of them?

python/metatomic_torch/metatomic/torch/dftd3.py
PicoCentauri: arbitrary but at least consistent within this implementation. I mean every energy split is arbitrary and kinda wrong.

GardevoirX: I see, but what's the possible use case of having this corrected atomic energy?

Luthaf: Same use case as someone wanting the atomic energy. The main use case I've seen for this is caching the energy and only recomputing parts of it in Monte-Carlo simulations, where only a few atoms moved.

python/metatomic_torch/metatomic/torch/dftd3.py
Luthaf: Yes, I'd rather make this explicit

python/metatomic_torch/metatomic/torch/dftd3.py
```diff
@@ -87,10 +88,13 @@ class DFTD3(torch.nn.Module):
...
-    independent. Damping parameters (``a1``, ``a2``, ``s8``, ...) are
-    provided per variant. All D3 tables and damping parameters **must** be
-    passed in **atomic units**. The wrapper converts the final D3 energy into the
-    wrapped model's energy unit of the corresponding output.
+    independent. The reference tables include four sets of parameters, ``rcov``
+    (dimension of length), ``r4r2`` (dimension of length), ``c6`` (dimension of energy *
+    length^6), and ``cn_ref`` (dimensionless). Damping parameters (``a1``, ``a2``,
+    ``s8``, ...) are provided per variant. Among these damping parameters, only ``a2``
+    has a dimension of length, the others are dimensionless. All D3 tables and damping
+    parameters **must** be passed in **atomic units**. The wrapper converts the final D3
```
Luthaf: This still does not say what the unit for lenght & energy should be

Luthaf (approved):

PR #223, opened 2026-05-07
Add spin/charge to ase calculator

JonathanSchmidt1:
We add spin and charge and corresponding tests to the ase calculator. Internally spin is called spin_multiplicity.
We also add a method to confirm that any new requested_inputs are not stale so that only updating these is sufficient for a recalculation of energies/forces.

# Reviewer checklist

### DIFF
--- a/docs/src/quantities/charge.rst
+++ b/docs/src/quantities/charge.rst
@@ -21,9 +21,12 @@ name (see :ref:`quantity-variants`), and must have the following metadata:
       :py:class:`metatensor.torch.TensorMap` with a single block.
 
   * - samples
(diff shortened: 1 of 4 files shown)


Luthaf (approved):

Luthaf:
I've made one small change, the code now looks for `atoms.info["spin_multiplicity"]` first, and then fallback to `atoms.info["spin"]`. IMO the "spin" fallback is mostly there for compatibility with existing dataset that use this name, but should not be the default one.

Otherwise this looks good, thanks for sticking with it @JonathanSchmidt1!

PR #245, opened 2026-05-29
Port unit parsing from metatomic-torch

Luthaf:


metatomic-core/src/units.rs
```diff
@@ -1,7 +1,722 @@
...
+    /// Length dimension
+    const LENGTH: Dimension = Dimension { length: 1, time: 0, mass: 0, electric_current: 0, temperature: 0 };
+    /// Time dimension
+    const TIME: Dimension = Dimension { length: 0, time: 1, mass: 0, electric_current: 0, temperature: 0 };
+    /// Mass dimension
+    const MASS: Dimension = Dimension { length: 0, time: 0, mass: 1, electric_current: 0, temperature: 0 };
+    /// Electric charge dimension (current × time)
+    const CHARGE: Dimension = Dimension { length: 0, time: 1, mass: 0, electric_current: 1, temperature: 0 };
+    /// Temperature dimension
+    const TEMPERATURE: Dimension = Dimension { length: 0, time: 0, mass: 0, electric_current: 0, temperature: 1 };
```
frostedoyster, line 35: Is it a conscious decision that all these units are missing amout of substance (unit `mol`)?

Luthaf: Yes, because we want `kJ/mol` to have the same dimension as `eV`

PR #119, opened 2025-11-18
Add `SymmetrizedModel`

ppegolo:
Add `SymmetrizedModel` wrapper to compute O(3) average and metrics for metatomic models.

# Contributor (creator of pull-request) checklist

 - [x] Tests updated (for new features and bugfixes)?
 - [x] Documentation updated (for new features)?
 - ~~[ ] Issue referenced (for PRs that solve an issue)?~~

# Reviewer checklist

 - [x] CHANGELOG updated with public API or any other important changes?

[📚 Download documentation for this pull-request](https://nightly.link/metatensor/metatomic/actions/artifacts/8348879471.zip)

[⚙️ Download Python wheels for this pull-request (you can install these with pip)](https://nightly.link/metatensor/metatomic/actions/artifacts/8349178862.zip)

PR #264, opened 2026-06-17
Implement augmentation routines in metatomic

ppegolo:
Augmentation is currently implemented in metatrain, and it depends on `spherical` for augmentation of spherical targets. 
In prepraration for the `SymmetrizedModel` PR (currently #119, that would depend on metatrain just for the augmentation routines) I re-implemented the relevant parts of [spherical](https://github.com/moble/spherical) (MIT license) here, getting rid of unnecessary stuff and most dependencies, and all the necessary functions that handle augmentation.

# Reviewer checklist

[📚 Download documentation for this pull-request](https://nightly.link/metatensor/metatomic/actions/artifacts/8166158224.zip)

[⚙️ Download Python wheels for this pull-request (you can install these with pip)](https://nightly.link/metatensor/metatomic/actions/artifacts/8166304299.zip)


python/metatomic_torch/metatomic/torch/_augmentation/_wigner.py
```diff
@@ -0,0 +1,527 @@
+"""Private Wigner-d/Wigner-D helpers for symmetry operations.
```
Luthaf: I wondering if this would be better inside https://github.com/Luthaf/wigners? but OK for now

python/metatomic_torch/metatomic/torch/o3/_wigner.py
Luthaf, line 1: maybe move this to `metateomic/torch/_augmentation/_wigner.py`

python/metatomic_torch/tests/augmentation.py
Luthaf, line 1: this is a bit hard to follow, would be nice to make it clearer/better documented

python/metatomic_torch/metatomic/torch/_augmentation.py
```diff
@@ -0,0 +1,545 @@
...
+import metatensor.torch as mts
+import numpy as np
+import torch
+from metatensor.torch import TensorBlock, TensorMap
+
+from . import System, register_autograd_neighbors
+from ._wigner import compute_wigner_batch
+
+
+def _block_row_indices_by_system(
```
Luthaf: Is this faster than doing `samples.select(Labels("system", [[i]]))`?

python/metatomic_torch/metatomic/torch/_augmentation.py
```diff
@@ -0,0 +1,545 @@
...
+        types=system.types,
+        cell=system.cell @ transformation.T,
+        pbc=system.pbc,
+    )
+    for data_name in system.known_data():
+        data = system.get_data(data_name)
+        if len(data) != 1:
+            raise ValueError(
+                f"System data '{data_name}' has {len(data)} blocks; "
+                "only single-block data is supported."
```
Luthaf: why?

python/metatomic_torch/metatomic/torch/_augmentation.py
```diff
@@ -0,0 +1,545 @@
...
+                            values=(
+                                block.values.swapaxes(-1, -2) @ transformation.T
+                            ).swapaxes(-1, -2),
+                            samples=block.samples,
+                            components=block.components,
+                            properties=block.properties,
+                        )
+                    ],
+                ),
+            )
```
Luthaf: This should use `_transform_tmap`

python/metatomic_torch/metatomic/torch/_augmentation.py
```diff
@@ -0,0 +1,545 @@
...
+    return list(R.unbind(0))
+
+
+def apply_augmentations(
+    systems: list[System],
+    targets: dict[str, TensorMap],
+    transformations: list[torch.Tensor],
+    extra_data: dict[str, TensorMap] | None = None,
+) -> tuple[list[System], dict[str, TensorMap], dict[str, TensorMap]]:
+    """Apply a batch of O(3) transformations to systems and TensorMaps simultaneously.
```
Luthaf: the transformations are rotation + inversion, right? I'm wondering if a better API would be expressed in term of transformations or rotations, instead of calling these "augmentations".

What you can do in the end is the same, just makes it a bit clearer when looking for functionality

Luthaf (commented):
Looks good overall!

python/metatomic_torch/metatomic/torch/_augmentation/_wigner.py
ppegolo: Maybe for later, yes, unless you think it's trivial?

python/metatomic_torch/metatomic/torch/_augmentation.py
ppegolo: No reason, now it's fine

python/metatomic_torch/metatomic/torch/_augmentation/_wigner.py
Luthaf: Trivial is a strong word, but I might give it a quick try. Otherwise we can merge this and change it later

Luthaf: It was not that bad, I think I can have a release by the end of the day https://github.com/Luthaf/wigners/pull/18

Luthaf: Can you check if the Python API makes sense for this use-case? https://github.com/Luthaf/wigners/pull/18/files#diff-c87cace15e30e245e07e40b204426e6d8246a567838786e0fb924f8ffc99fe93R96

ppegolo: Thank you! At a first glance I think it could substitute `_compute_wigner_d_complex`, so in the current state it should still be massaged to be used here, but it's probably fine. Not sure I have time to test it today, but I'll try.

ppegolo: Tested, it works!

python/metatomic_torch/metatomic/torch/_augmentation.py
ppegolo: Yes, a bit more than twice as fast

python/metatomic_torch/tests/augmentation.py
ppegolo: hopefully clearer now

Luthaf:
I think I'm happy with the new API, I'll let you check if this is fine for you as well @ppegolo. Main changes are 

- no augmentation/dataset handling, only `transform_{system,tensor,block}`
- new O3Transformation class to group the transformation matrix & wigner_D matrix cache
- the code no longer try to guess the system indices. They should either match `list(range(len(systems)))`, or be passed explicitly

Still TBD before merging:

- [x] add docstring for the different modules
- [x] Add some explanation in the docs of the convention we follow for everything (real sph, naming using `o3_`)

Luthaf (approved):

PR #267, opened 2026-06-23
C++ API for loading plugins

RMeli:

# Reviewer checklist

[📚 Download documentation for this pull-request](https://nightly.link/metatensor/metatomic/actions/artifacts/7851133293.zip)

[⚙️ Download Python wheels for this pull-request (you can install these with pip)](https://nightly.link/metatensor/metatomic/actions/artifacts/7846166789.zip)


metatomic-core/include/metatomic/plugin.hpp
```diff
@@ -1,7 +1,59 @@
...
+#include <metatomic/errors.hpp>
 
 namespace metatomic {
+    /// Register a plugin with metatomic.
+    ///
+    /// @param plugin the plugin to register
+    inline void register_plugin(mta_plugin_t plugin) {
+        auto status = mta_register_plugin(plugin);
+        details::check_status(status);
+    }
```
Luthaf: I don't know if we want to expose this one, given that even the C function should not really be called directly but rather through the macro.

metatomic-core/tests/plugins.cpp
```diff
@@ -1,45 +1,67 @@
...
+        CHECK(std::string(error_origin) == "metatomic-core");
+        expected_message = (
+            "invalid parameter: can not register plugin 'bad-abi-plugin': "
+            "plugin ABI version is 2, but metatomic expects 1"
+        );
+        CHECK(std::string(error_message) == expected_message);
+    }
+
+    SECTION("C++ API") {
+        REQUIRE_THROWS_AS(
```
Luthaf: Can you check the error message here as well, like below

metatomic-core/include/metatomic/plugin.hpp
```diff
@@ -1,7 +1,59 @@
...
+    /// @param plugin_name optional name of the plugin to use for loading the
+    ///     model, or empty to let metatomic search
+    /// @param options_json optional JSON object containing string keys and
+    ///     string values for loading the model
+    /// @return the loaded model
+    inline mta_model_t load_model(
+        const std::string& load_from,
+        const std::string& plugin_name = "",
+        const std::string& options_json = ""
+    ) {
```
RMeli, line 41: I did it this way to use default arguments; however, the order is different from the C API. Should we change the C API to the same order?

Luthaf: Yes, would make sense. In the order people are most to least likely to pass these I would even do `load_from, options_json, plugin_name`

RMeli: Sounds good!

metatomic-core/tests/plugins.cpp
RMeli: The error was not properly handled. I added a capture for `Error::CallbackError` and return `Error:InvalidParameter` instead with a clearer error message. Is this what you had in mind?

Luthaf: Yes, looks good

Luthaf:
I rebased this branch, it should be good to go if CI passes

PR #274, opened 2026-06-30
Custom cuda and metal kernels for system validation

Luthaf:
Add custom cuda and metal kernels for system validation. The intention is to add more kernels for unit conversion & any other work that needs to be done inside `mta_execute_model`.

No tests other than making sure the code compile for now, they will be added once we have Python bindings.

 - [ ] ~Tests updated (for new features and bugfixes)?~

# Reviewer checklist

### DIFF
--- a/metatomic-core/Cargo.toml
+++ b/metatomic-core/Cargo.toml
@@ -19,8 +19,13 @@ dlpk = { version = "0.4", features = ["ndarray"]}
 json = "0.12"
 libloading = "0.9"
 ndarray = "0.17"
+
+# For serialization of the systems
 zip = { version = "8.6.0", default-features = false }
 byteorder = {version = "1"}
+
+# For custom kernels
+cudarc = {version = "0.19", default-features = false, features=["std", "cuda-13030", "driver", "nvrtc", "dynamic-loading"]}
 
 
 [build-dependencies]
--- a/metatomic-core/Cargo.toml
+++ b/metatomic-core/Cargo.toml
@@ -27,6 +27,10 @@ byteorder = {version = "1"}
 # For custom kernels
 cudarc = {version = "0.19", default-features = false, features=["std", "cuda-13030", "driver", "nvrtc", "dynamic-loading"]}
 
+[target.'cfg(target_os = "macos")'.dependencies]
+objc2-metal = "0.3"
+objc2 = "0.6"
+objc2-foundation = "0.3"
 
 [build-dependencies]
 cbindgen = { version = "0.29", default-features = false }
--- a/metatomic-core/cmake/metatomic-config.in.cmake
+++ b/metatomic-core/cmake/metatomic-config.in.cmake
@@ -78,7 +78,15 @@ if (@METATOMIC_INSTALL_BOTH_STATIC_SHARED@ OR NOT @BUILD_SHARED_LIBS@)
     )
 
     target_compile_features(metatomic::static INTERFACE cxx_std_17)
-    target_link_libraries(metatomic::static INTERFACE metatensor nlohmann_json::nlohmann_json)
+
+    target_link_libraries(metatomic::static INTERFACE metatensor)
+    target_link_libraries(metatomic::static INTERFACE nlohmann_json::nlohmann_json)
+
+    if(APPLE)
+        target_link_libraries(metatomic::static INTERFACE
+            "-framework Metal" "-framework CoreGraphics" "-framework CoreFoundation" "-framework Foundation" objc
+        )
+    endif()
 endif()
 
 # Export either the shared or static library as the metatomic target
--- a/metatomic-core/src/kernels/cpu.rs
+++ b/metatomic-core/src/kernels/cpu.rs
@@ -1,0 +1,60 @@
+use dlpk::DLPackTensorRef;
+use ndarray::{ArrayView1, ArrayView2, ArrayViewD};
+
+use crate::Error;
+
+/// Check that the values of an i32 DLPack tensor match the expected reference.
+///
+/// The tensor is converted to an ndarray view and compared element-wise and
+/// shape-wise against `reference`. The `description` is used verbatim in the
+/// error message on mismatch.
+///
+/// # Parameters
+/// - `tensor`: DLPack tensor with i32 data type
+/// - `reference`: expected values with the same shape as the tensor
+pub(crate) fn is_equal_i32(
+    tensor: DLPackTensorRef<'_>,
+    reference: ArrayViewD<'_, i32>,
+) -> Result<bool, Error> {
+    let values: ArrayViewD<i32> = tensor.try_into()?;
+    return Ok(values == reference);
+}
+
+macro_rules! validate_cell {
+    ($T: ty, $pbc: expr, $cell: expr) => {
+        let pbc_array: ArrayView1<bool> = $pbc.try_into()?;
+        let cell_array: ArrayView2<$T> = $cell.try_into()?;
+        for i in 0..3 {
+            if !pbc_array[i] && !cell_array.row(i).iter().all(|&x| x == 0.0) {
+                return Err(Error::InvalidParameter(format!(
+                    "invalid cell: for non-periodic dimensions, the corresponding \
+                    cell vector must be zero, but cell[{}] contains non-zero values",
+                    i
+                )));
+            }
+        }
+    };
+}
+
+/// Validate that cell vectors are zero for non-periodic dimensions on CPU.
+///
+/// Converts the DLPack tensors to ndarray views and checks that for every
+/// dimension where `pbc` is false, the corresponding row of `cell` contains
+/// only zeros.
+///
+/// # Parameters
+/// - `pbc`: 1D boolean tensor of length 3 (periodic boundary condition flags)
+/// - `cell`: 3x3 tensor (unit cell vectors as rows)
+pub(crate) fn validate_cell_pbc(
+    pbc: DLPackTensorRef<'_>,
+    cell: DLPackTensorRef<'_>,
+) -> Result<(), Error> {
+    let dtype = cell.dtype();
+    if dtype.bits == 32 {
+        validate_cell!(f32, pbc, cell);
+    } else {
+        assert_eq!(dtype.bits, 64);
+        validate_cell!(f64, pbc, cell);
+    }
+    return Ok(());
+}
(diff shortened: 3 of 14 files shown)


metatomic-core/src/kernels/cpu.rs
```diff
@@ -0,0 +1,69 @@
...
+        for i in 0..3 {
+            if !pbc_array[i] && !cell_array.row(i).iter().all(|&x| x == 0.0) {
+                return Err(Error::InvalidParameter(format!(
+                    "invalid cell: for non-periodic dimensions, the corresponding \
+                    cell vector must be zero, but cell[{}] contains non-zero values",
+                    i
+                )));
+            }
+        }
+    } else {
```
RMeli, line 56: Worth checking for `dtype.bits==64` explicitly and error out in all other cases?

metatomic-core/cmake/metatomic-config.in.cmake
```diff
@@ -78,7 +78,15 @@ if (@METATOMIC_INSTALL_BOTH_STATIC_SHARED@ OR NOT @BUILD_SHARED_LIBS@)
...
     target_compile_features(metatomic::static INTERFACE cxx_std_17)
-    target_link_libraries(metatomic::static INTERFACE metatensor nlohmann_json::nlohmann_json)
+
+    target_link_libraries(metatomic::static INTERFACE metatensor)
+    target_link_libraries(metatomic::static INTERFACE nlohmann_json::nlohmann_json)
+
+    if(APPLE)
+        target_link_libraries(metatomic::static INTERFACE
+            "-framework Metal" "-framework CoreGraphics" "-framework CoreFoundation" objc
+        )
```
RMeli, line 88: Only `metatomic::static` target?

metatomic-core/Cargo.toml
```diff
@@ -19,9 +19,17 @@ dlpk = { version = "0.4", features = ["ndarray"]}
 json = "0.12"
 libloading = "0.9"
 ndarray = "0.17"
+
+# For serialization of the systems
 zip = { version = "8.6.0", default-features = false }
 byteorder = {version = "1"}
 
+# For custom kernels
+cudarc = {version = "0.19", default-features = false, features=["std", "cuda-13030", "driver", "nvrtc", "dynamic-loading"]}
```
RMeli, line 28: Is this always pulled as a dependency?

metatomic-core/src/kernels/cpu.rs
```diff
@@ -0,0 +1,69 @@
...
+        let cell_array: ArrayView2<f64> = cell.try_into()?;
+        for i in 0..3 {
+            if !pbc_array[i] && !cell_array.row(i).iter().all(|&x| x == 0.0) {
+                return Err(Error::InvalidParameter(format!(
+                    "invalid cell: for non-periodic dimensions, the corresponding \
+                    cell vector must be zero, but cell[{}] contains non-zero values",
+                    i
+                )));
+            }
+        }
```
RMeli: This is essentially the same as above, but with a different type for `ArrayView2`. An helper function would avoid code duplication (and make it easier to add support to more exotic types down the line if needed).

metatomic-core/src/kernels/cuda.rs
```diff
@@ -0,0 +1,227 @@
...
+    let mut result = stream.alloc_zeros::<i32>(1)
+        .map_err(|e| Error::Internal(format!("alloc_zeros: {e}")))?;
+
+    unsafe {
+        stream.launch_builder(&entry.is_equal_i32)
+            .arg(&tensor_ptr)
+            .arg(&ref_dev)
+            .arg(&n)
+            .arg(&mut result)
+            .launch(LaunchConfig::for_num_elems(n as u32))
```
RMeli: `n` is defined as `i64`. I guess this truncation is fine, but might be worth adding a check that `n` is less than whatever the Rust equivalent of `std::numeric_limits<std::uint32_t>::max()` is?

metatomic-core/src/kernels/cuda_kernels.cu
```diff
@@ -0,0 +1,43 @@
...
+////////////////////////////////////////////////////////////////////////////////
+
+template <typename T>
+__device__ void validate_cell_pbc_impl(
+    const unsigned char* pbc, const T* cell, int* mismatch_idx
+) {
+    int i = threadIdx.x;
+    if (i < 3) {
+        if (!pbc[i]) {
+            if (cell[3 * i] != T(0) || cell[3 * i + 1] != T(0) || cell[3 * i + 2] != T(0)) {
```
RMeli: Given this indexing, worth checking that DLPack tensors are contiguous in memory before calling the kernel? If I understand correctly, they could potentially be strided in general.

metatomic-core/src/kernels/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+                    "Metal backend is only available on macOS".into(),
+                ))
+            }
+        }
+        _ => {
+            eprintln!(
+                "is_equal_i32 for device {:?} is not implemented",
+                tensor.device()
+            );
+            Ok(())
```
RMeli: Shouldn't this be an error?

metatomic-core/src/kernels/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+                    "Metal backend is only available on macOS".into(),
+                ))
+            }
+        }
+        _ => {
+            eprintln!(
+                "Cell/PBC validation for device {:?} is not implemented",
+                pbc.device()
+            );
+            Ok(())
```
RMeli, line 165: As above, shouldn't this be an error?

metatomic-core/cmake/metatomic-config.in.cmake
Luthaf: Yes, for the shared one cargo is already linking the correct libraries and adding them to the DT_NEEDED equivalent

metatomic-core/Cargo.toml
Luthaf: Yes, it works exactly like gpulite and fails at runtime with symbols not found if cuda is not actually available. I prefer something like this, since it means we will only need to build and distribute one version of the code.

metatomic-core/src/kernels/cuda_kernels.cu
Luthaf: very good point!

metatomic-core/src/kernels/mod.rs
Luthaf: I don't know since these are only checks that don't change the model behavior, I would rather skip the checks than error here, to make it easier to add new devices in metatomic without having to add code everywhere

metatomic-core/src/kernels/cpu.rs
Luthaf: fixed

metatomic-core/src/kernels/cpu.rs
Luthaf: fixed with a macro (rust generics don't support unconstrained "template parameters", they always require "concepts" i.e. traits for all parameters)

metatomic-core/src/kernels/cuda.rs
Luthaf: fixed with `try_from`

metatomic-core/src/kernels/cuda_kernels.cu
Luthaf: This was quite painful and added a bunch of code, but should be fine now

metatomic-core/src/kernels/mod.rs
RMeli: Ok, fair enough.

metatomic-core/src/kernels/cuda.rs
```diff
@@ -0,0 +1,245 @@
...
+                .arg(&cell_idx)
+                .arg(&mut result)
+                .launch(LaunchConfig {
+                    grid_dim: (1, 1, 1),
+                    block_dim: (3, 1, 1),
+                    shared_mem_bytes: 0,
+                })
+                .map_err(|e| Error::Internal(format!("kernel launch (f32): {e}")))?;
+        }
+    } else {
```
RMeli, line 207: I missed this on the first pass, worth checking `cell.dtype().bits == 64` explicitly?

metatomic-core/src/kernels/metal.rs
```diff
@@ -0,0 +1,237 @@
...
+fn get_or_init(device_id: usize) -> Result<&'static MetalKernelCache, Error> {
+    let mut cache = METAL_CACHE.lock().unwrap();
+    let entry = match cache.entry(device_id) {
+        std::collections::hash_map::Entry::Occupied(e) => e.into_mut(),
+        std::collections::hash_map::Entry::Vacant(e) => {
+            e.insert(MetalKernelCache::new(device_id)?)
+        }
+    };
+    // Safety: the cache lives for the lifetime of the process (Lazy static),
+    // so the reference is valid for 'static once initialized and never removed.
```
RMeli: I'm not very familiar with Rust, but from the documentation looks like `HashMap` can re-allocate. Is this taken into account?

metatomic-core/src/kernels/metal.rs
```diff
@@ -0,0 +1,237 @@
...
+
+    // Build strided indices for values and reference
+    let values_idx = StridedNDIndex::from_dlpack(&tensor);
+    let reference_idx = StridedNDIndex::from_ndarray(&reference);
+
+    let opts = MTLResourceOptions::CPUCacheModeDefaultCache;
+
+    let values_buf = entry.device.new_buffer_with_data(values_ptr, values_bytes, opts);
+    let ref_buf = entry.device.new_buffer_with_data(ref_ptr, ref_bytes, opts);
+    let result_buf = entry.device.new_buffer(std::mem::size_of::<i32>() as u64, opts);
```
RMeli: In the CUDA version, the result buffer is initialized to 0. Here it doesn't seem to be initialized, is it?

RMeli (commented):
A few final comments, but probably missed some other things without tests... We can revisit once the Python bindings are available.

metatomic-core/src/kernels/metal.rs
Luthaf: great catch, thanks!

metatomic-core/src/kernels/metal.rs
Luthaf: yep, we don't actually need a `'static` value here, I simplified it!

metatomic-core/src/kernels/cuda.rs
Luthaf: done

PR #277, opened 2026-07-02
C++ classes with JSON serialization/deserialization

RMeli:
This PR add C++ classes matching what is available in Rust for convenience, allowing automated JSON serialization/deserialization.

# Reviewer checklist

[📚 Download documentation for this pull-request](https://nightly.link/metatensor/metatomic/actions/artifacts/8143036115.zip)

[⚙️ Download Python wheels for this pull-request (you can install these with pip)](https://nightly.link/metatensor/metatomic/actions/artifacts/8143255953.zip)


Luthaf:
I had to rebase this branch, you can use the following to update your local branch instead of `git pull`

```
git checkout rmeli-cpp-json
git fetch upstream # or however this remote is named
git reset --hard upstream/rmeli-cpp-json
```

metatomic-core/CMakeLists.txt
```diff
@@ -499,5 +502,6 @@ endif()
 install(FILES
     ${PROJECT_BINARY_DIR}/metatomic-config-version.cmake
     ${PROJECT_BINARY_DIR}/metatomic-config.cmake
+    ${PROJECT_SOURCE_DIR}/cmake/nlohmann_json.cmake
```
Luthaf: Why install a cmake script to download the lib instead of installing the headers in `include/metatomic/third_party/nlohmann/`?

metatomic-core/tests/misc.cpp
```diff
@@ -5,16 +5,16 @@
...
+//     auto version = std::to_string(METATOMIC_VERSION_MAJOR) + "."
+//         + std::to_string(METATOMIC_VERSION_MINOR) + "."
+//         + std::to_string(METATOMIC_VERSION_PATCH);
 
-    // METATOMIC_VERSION should start with `x.y.z`
-    CHECK(std::string(METATOMIC_VERSION).find(version) == 0);
-}
+//     // METATOMIC_VERSION should start with `x.y.z`
+//     CHECK(std::string(METATOMIC_VERSION).find(version) == 0);
+// }
```
Luthaf: Can you re-enable this test?

metatomic-core/include/metatomic/metadata.hpp
```diff
@@ -0,0 +1,846 @@
...
+#include <nlohmann/json.hpp>
+
+namespace metatomic{
+    namespace detail {
+
+    inline std::vector<std::string> read_string_array(
+        const nlohmann::json& j, const std::string& key, const std::string& context
+    ) {
+        if (!j.contains(key) || !j[key].is_array()) {
+            throw metatomic::Error("'" + key + "' " + context + " must be an array");
```
Luthaf: small nit: I would put the "in" here instead of inside the context

```suggestion
            throw metatomic::Error("'" + key + "' in " + context + " must be an array");
```

metatomic-core/include/metatomic/metadata.hpp
```diff
@@ -0,0 +1,846 @@
...
+        for (char c : s) {
+            if (!(is_ascii_alphanumeric(c) || c == '_')) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    inline void validate_quantity_name(const std::string& name) {
```
Luthaf: I don't know if we want to redo this validation here, worst case the failure on invalid name happens a bit later when passing data to the C API.

metatomic-core/include/metatomic/metadata.hpp
```diff
@@ -0,0 +1,846 @@
...
+
+        PairListOptions(double cutoff_, bool full_list_, bool strict_, const std::vector<std::string>& requestors_  = {})
+            : cutoff(cutoff_), full_list(full_list_), strict(strict_), requestors(requestors_) {
+            if (!std::isfinite(cutoff_) || cutoff_ <= 0.0) {
+                throw metatomic::Error("cutoff must be a finite positive number");
+            }
+        }
+    };
+
+    inline void to_json(nlohmann::json& j, const PairListOptions& p){
```
Luthaf, line 166: Why not make this a member function? Both for `to_json` and `from_json`?

Luthaf (commented):
The different fields in the new struct should have some documentation, and the new structs should be added to the sphinx documentation as well

metatomic-core/tests/cxx/metadata.cpp
Luthaf, line 1: can you move this to `tests/cxx`?

Luthaf:
The build error on windows is 

```
CUSTOMBUILD : error : unable to create file tests/reports/2016-09-09-nativejson_benchmark/performance_Corei7-4980HQ@2.80GHz_mac64_clang7.0_7._Code_size_FileSize_(byte).png: 
Filename too long [D:\a\metatomic\metatomic\target\x86_64-pc-windows-msvc\tmp\cxx-install\cmake-find-package\deps\metatomic-core\_deps\nlohmann_json-subbuild\nlohmann_json-populate.vcxproj]
```

Maybe we should try to enable long path on CI? Or maybe there is a way to get an archive of the code that does not include this but only has the headers?

metatomic-core/tests/misc.cpp
RMeli: Yep! This test fails for me when I'm developing, but I'm not yet sure why.

metatomic-core/include/metatomic/metadata.hpp
RMeli: I wasn't sure either. I would prefer as little duplication as possible, but added it anyways for the time being to throw.

I also used the C API for checking that `length_unit` is correct (as in Rust), but since `validate_units` is not exposed in the C API, I used `mta_unit_conversion_factor` instead. Do you prefer to avoid that validation too in the C++ API?

metatomic-core/CMakeLists.txt
RMeli: The script does not download the library unconditionally; it searches if it is already available (this is important for packaging thing properly). I'll see if there is a more elegant solution for supporting both use cases.

metatomic-core/include/metatomic/metadata.hpp
RMeli: This is how the library works; it uses argument-dependent lookup to locate those functions. Since I deleted the default constructors (to match Rust) it is even a bit more complicated: the `adl_serializer` needs to be specialized for our non-default-constructable types and injected in the `nlohmann` namespace.

metatomic-core/include/metatomic/metadata.hpp
RMeli: I removed `validate_quantity_name` and related functions/uses.

metatomic-core/CMakeLists.txt
RMeli: Now I install the JSON library when a system library is not found, so that should be cleaner.

RMeli:
As discussed offline, I did some refactoring allowing to use a default constructor while still forcing the user to initialize all fields (using AI).

RMeli:
> Or maybe there is a way to get an archive of the code that does not include this but only has the headers?

I went with downloading the archive, which is also the recommended method on their README.

metatomic-core/cmake/nlohmann_json.cmake
```diff
@@ -0,0 +1,35 @@
...
+  # but not the benchmark reports with very long filenames that break Windows.
+  FetchContent_Declare(
+    nlohmann_json
+    URL https://github.com/nlohmann/json/releases/download/v3.11.3/json.tar.xz
+  )
+
+  set(JSON_BuildTests OFF CACHE INTERNAL "")
+  set(JSON_Install ON CACHE INTERNAL "")
+
+  FetchContent_MakeAvailable(nlohmann_json)
```
Luthaf, line 32: Trying to make sure I understand everything. This will install the library when doing `make install` in such a way that it can later be found by `metatomic-config.cmake`? In a scenario where someone first configure-build-install metatomic; and then tries to configure a package that depends on metatomic cmake and have it use the installed version?

metatomic-core/include/metatomic/metadata.hpp
```diff
@@ -0,0 +1,949 @@
...
+
+        /// Get whether this pair list is strict.
+        bool strict() const {
+            return strict_;
+        }
+
+        /// Set the list of requestors for this pair list.
+        void requestors(const std::vector<std::string>& value) {
+            requestors_ = value;
+        }
```
Luthaf, line 106: Should we do `add_requestor`/`clear_requestors` instead of just a setter that overwrite everything?

metatomic-core/include/metatomic/metadata.hpp
```diff
@@ -0,0 +1,949 @@
...
+        PairListOptions(
+            double cutoff_,
+            bool full_list_,
+            bool strict_ = true,
+            const std::vector<std::string>& requestors_ = {}
+        ) {
+            cutoff(cutoff_);
+            full_list(full_list_);
+            strict(strict_);
+            requestors(requestors_);
```
Luthaf:
```suggestion
            this->cutoff(cutoff_);
            this->full_list(full_list_);
            this->strict(strict_);
            this->requestors(requestors_);
```

For clarity, because I was confused for a second here

metatomic-core/include/metatomic/metadata.hpp
```diff
@@ -0,0 +1,949 @@
...
+            std::vector<std::string> model_;
+            /// The references about the architecture of the model, e.g. papers
+            /// describing the mathematical form of the model.
+            std::vector<std::string> architecture_;
+            /// The references about the implementation of the model, e.g. a link to
+            /// the source code repository or a paper describing the software.
+            std::vector<std::string> implementation_;
+
+        public:
+            void model(const std::vector<std::string>& value) {
```
Luthaf: same comment about having `add_` and `clear_`  function instead of/in addition to this one

metatomic-core/include/metatomic/metadata.hpp
```diff
@@ -0,0 +1,949 @@
...
+    private:
+        std::string name_;
+        std::vector<std::string> authors_;
+        std::string description_;
+        References references_;
+        // BTreeMap in Rust is an ordered map
+        std::map<std::string, std::string> extra_;
+
+    public:
+        void name(const std::string& value) {
```
Luthaf: Can this get some documentation?

metatomic-core/include/metatomic/metadata.hpp
```diff
@@ -0,0 +1,949 @@
...
+            /// The references about the implementation of the model, e.g. a link to
+            /// the source code repository or a paper describing the software.
+            std::vector<std::string> implementation_;
+
+        public:
+            void model(const std::vector<std::string>& value) {
+                model_ = value;
+            }
+
+            const std::vector<std::string>& model() const {
```
Luthaf, line 279: These all need some documentation

metatomic-core/include/metatomic/metadata.hpp
```diff
@@ -0,0 +1,949 @@
...
+    public:
+        void name(const std::string& value) {
+            name_ = value;
+        }
+
+        const std::string& name() const {
+            return name_;
+        }
+
+        void authors(const std::vector<std::string>& value) {
```
Luthaf: Setters that take string/vectors should take them by value and move them in the function member

metatomic-core/include/metatomic/metadata.hpp
```diff
@@ -0,0 +1,949 @@
...
+        const std::map<std::string, std::string>& extra() const {
+            return extra_;
+        }
+
+        ModelMetadata(
+            const std::string& name_ = "",
+            const std::vector<std::string>& authors_ = {},
+            const std::string& description_ = "",
+            const References& references_ = {},
+            const std::map<std::string, std::string>& extra_ = {}
```
Luthaf: All these should be passed by value and then moved

metatomic-core/include/metatomic/metadata.hpp
```diff
@@ -0,0 +1,949 @@
...
+
+        /// The gradients a quantity can have
+        enum class Gradients {
+            /// Gradients with respect to atomic positions
+            Positions,
+            /// Gradients with respect to the strain (typically used for stress)
+            Strain,
+        };
+
+        /// A quantity that a model can use as input or output
```
Luthaf: Can you move quantity to a top level struct, and not one scoped inside ModelCapabilities? It can be used by itself (for example when declaring inputs/outputs of a model.

metatomic-core/include/metatomic/metadata.hpp
RMeli: Good point!

metatomic-core/cmake/nlohmann_json.cmake
RMeli: Yes, if `nlohmann_json` is downloaded and built by `metatomic-core` it gets installed under `metatomic-core/usr/include/nlohmann` and it is correctly found by `find_dependency(nlohmann_json 3.11.0)` in `metatomic-config.cmake`.

If it is found externally (provided by the system, provided by Spack, etc), then `find_dependency(nlohmann_json 3.11.0)` will also need to find the external one when configuring a package that depends on `metatomic`.

What we might add (useful for Spack), is a way to force one behavior or the other, something like `METATOMIC_FORCE_EXTERNAL_JSON`/`METATOMIC_FORCE_INTERNAL_JSON`.

Luthaf: Makes sense. Maybe `METATOMIC_DOWNLOAD_JSON`, which default to ON if we can not find `nlohmann_json`? Although I remember hitting some issue with the fact that `find_package(nlohmann_json)` would create the corresponding targets, and if the user then forces downloading anyway the second target creation would fail

Luthaf: ok, I kinda need this branch to finish the System API, so I'll merge in the current state and we can add this option as a follow up!

PR #278, opened 2026-07-03
Support explicit_gradients in run_model to compute energy_ensemble Jacobians

ppegolo:
Support models that can compute ensemble forces and stresses with `explicit_gradients`.

Added also some tests that implement artificial gradients to check functionality.

# Reviewer checklist

### DIFF
--- a/python/metatomic_ase/CHANGELOG.md
+++ b/python/metatomic_ase/CHANGELOG.md
@@ -6,9 +6,19 @@ follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased](https://github.com/metatensor/metatomic/)
 
-<!-- Possible sections
 ### Added
 
+- `MetatomicCalculator` now resolves and exposes an `energy_ensemble` output
+  (respecting `variants`), like it already does for `energy` and
+  `energy_uncertainty`.
+- `MetatomicCalculator.run_model()` now makes `positions` and `cell` require grad
+  before building the `System` whenever any requested output has a non-empty
+  `explicit_gradients`, so models that compute their own gradients (e.g. an
+  ensemble of forces/stress alongside `energy_ensemble`) can do so through the
+  standard `run_model()` call, without any extra autograd handling on the
+  caller's side.
+
+<!-- Possible sections
 ### Fixed
 
 ### Changed
--- a/python/metatomic_ase/src/metatomic_ase/_calculator.py
+++ b/python/metatomic_ase/src/metatomic_ase/_calculator.py
@@ -212,10 +212,10 @@ class MetatomicCalculator(ase.calculators.calculator.Calculator):
             the options in the model's ``supported_device`` in order.
         :param variants: dictionary mapping output names to a variant that should be
             used for the calculations (e.g. ``{"energy": "PBE"}``). If ``"energy"`` is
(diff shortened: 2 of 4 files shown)


Luthaf:
I don't love that this is specific to ASE, and that we are adding more special purpose functions. The end goal is to have force/stress uncertainty, correct? Or do you really want to ensemble of forces/stress to be available? 

This does not support non-conservative forces/stress either?

ppegolo:
The goal is to have ensembles, to be able to propagate the uncertainty more easily to downstream properties (e.g. phonons; for stress I don't necessarily have a use-case in mind). At this stage it doesn't include non conservative stuff, but I can can add it (this started as a PR to `upet`, where that's included, but the conservative path involved private `metatomic_ase` functions, so it probably fits better here in metatomic). Would you prefer it at a lower level? Or you don't like the idea of having it at all?

Luthaf:
I think we should have a way to do this, but I'd rather not have a specific function just for this use case!

This looks like a perfect use case for setting [`explicit_gradients`](https://docs.metatensor.org/metatomic/latest/torch/reference/models/metadata.html#metatomic.torch.ModelOutput.explicit_gradients) on ModelOutput, and asking the model to compute gradients itself. Then PET (or some wrapper) can call `torch.autograd.functional.jacobian` internally to compute these. This would also work in the same way in i-PI or anyone else who wants the full ensemble of forces.

Another reason not to do it inside metatomic_ase is that once we move to the C API, metatomic_ase will no longer have access to torch and this code will need to go away anyway.

python/metatomic_ase/CHANGELOG.md
```diff
@@ -6,9 +6,19 @@ follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
...
 
+- `MetatomicCalculator` now resolves and exposes an `energy_ensemble` output
+  (respecting `variants`), like it already does for `energy` and
+  `energy_uncertainty`.
+- `MetatomicCalculator.run_model()` now makes `positions` and `cell` require grad
+  before building the `System` whenever any requested output has a non-empty
+  `explicit_gradients`, so models that compute their own gradients (e.g. an
+  ensemble of forces/stress alongside `energy_ensemble`) can do so through the
+  standard `run_model()` call, without any extra autograd handling on the
+  caller's side.
```
Luthaf, line 19: This is not the best solution here IMO. Once we move away from torch this won't be possible, so we can as well have the code inside LLPR set requires_grad explicitly if they want to use torch autograd jacobian.

python/metatomic_ase/CHANGELOG.md
```diff
@@ -6,9 +6,19 @@ follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased](https://github.com/metatensor/metatomic/)
 
-<!-- Possible sections
 ### Added
 
+- `MetatomicCalculator` now resolves and exposes an `energy_ensemble` output
+  (respecting `variants`), like it already does for `energy` and
+  `energy_uncertainty`.
```
Luthaf, line 13: I don't really understand what this means. ASE has no concept of energy_ensemble, so what does resolving and exposing entails?

ppegolo:
Closing as not needed: only changes to metatrain are required

PR #279, opened 2026-07-05
Implement System in the C++ API

frostedoyster:
Also includes small helpers for strings and DLPack

 - ~Issue referenced (for PRs that solve an issue)?~

# Reviewer checklist

 - ~CHANGELOG updated with public API or any other important changes?~

[📚 Download documentation for this pull-request](https://nightly.link/metatensor/metatomic/actions/artifacts/8193067559.zip)

[⚙️ Download Python wheels for this pull-request (you can install these with pip)](https://nightly.link/metatensor/metatomic/actions/artifacts/8095461666.zip)

### DIFF
--- a/metatomic-core/include/metatomic/system.hpp
+++ b/metatomic-core/include/metatomic/system.hpp
@@ -1,7 +1,320 @@
 #pragma once
 
+#include <cstddef>
+#include <string>
+#include <utility>
+
 #include <metatomic.h>
+#include <metatensor.hpp>
+
+#include <metatomic/errors.hpp>
+#include <metatomic/utils.hpp>
+#include <metatomic/metadata.hpp>
 
 namespace metatomic {
+    /// A `System` contains all the information about an atomistic system, and is
+    /// used as the input of atomistic models.
+    ///
+    /// This is a RAII wrapper around the `mta_system_t` type from the C API. It
+    /// can either own the underlying system (in which case it is freed with the
+    /// `System`), or be a non-owning view into a system owned elsewhere (for
+    /// example a system passed to a model by the runtime).
+    class System final {
+    public:
+        /// Create a new `System` from DLPack tensors.
+        ///
+        /// Ownership of all four tensors is transferred to the new `System`.
+        ///
+        /// @param length_unit unit of length used by `positions` and `cell`
+        /// @param types tensor with shape `(n_atoms,)` of atomic types
+        /// @param positions tensor with shape `(n_atoms, 3)` of atomic positions
+        /// @param cell tensor with shape `(3, 3)` of the unit cell vectors
+        /// @param pbc tensor with shape `(3,)` of periodic boundary conditions
+        ///
+        /// The dtype and layout required for each tensor are validated by
+        /// `mta_system_create`; see the C API documentation for details.
+        System(
+            const std::string& length_unit,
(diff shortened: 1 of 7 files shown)


RMeli (commented):
LGTM. I'd consider having `System` and `SystemView` instead of the current `is_view_` mechanism, which I think would be cleaner.

frostedoyster:
@Luthaf what would you say regarding the views as a separate class?

Luthaf:
We went kinda back and forth with this. We initially had a `LabelsView` class that was removed, and now `TensorMap`/`TensorBlock` have a `is_view_` field. 

Overall I have a small preference for `is_view_`, because this makes writing functions that take `System` as input easier. Otherwise we have to either pick one of System/SystemView and force the caller to use `to_view()` half of the time; or make the function a template.

EDIT: Although because this is C++, maybe `SystemView operator SystemView()` on `System` can help, and then all functions that just need a view can be defined taking `SystemView` and automatically work with `System` as well.

RMeli:
> Although because this is C++, maybe SystemView operator SystemView() on System can help, and then all functions that just need a view can be defined taking SystemView and automatically work with System as well.

Yes, this is what I had in mind; System could be implicitly converted to StringView.

Anyways, mine was just a suggestion and certainly non-blocking. If you have a preference for `is_view_` and already went back and forth with the design decision, I don't mind either approach.

Luthaf:
I updated the code to use the new types from #277, It should be good to merge if CI is happy

metatomic-core/include/metatomic/system.hpp
```diff
@@ -1,7 +1,322 @@
...
+
+        /// Add a pair list (i.e. neighbor list) to this system.
+        ///
+        /// Ownership of `pairs` is transferred to this `System`.
+        ///
+        /// @param options options describing the pair list
+        /// @param pairs pairs data, stored as a metatensor block
+        void add_pairs(const PairListOptions& options, metatensor::TensorBlock pairs) {
+            nlohmann::json j;
+            to_json(j, options);
```
RMeli:
```suggestion
            nlohmann::json j = options;
```
The `to_json` function should be automatically detected and called. The `add_pairs` option could even take `nlohmann::json` as input so that the end user can decide if they want to pass a JSON directly or a `PairListOptions`, but maybe the documentation would be less clear with such signature.

metatomic-core/include/metatomic/system.hpp
```diff
@@ -1,7 +1,322 @@
...
+
+        /// Get a previously stored pair list matching the given `options_json`.
+        ///
+        /// The returned block is a non-owning view into data owned by this
+        /// `System`, and is only valid for as long as this `System` is alive.
+        ///
+        /// @param options options identifying the pair list to retrieve
+        metatensor::TensorBlock pairs(const PairListOptions& options) const {
+            nlohmann::json j;
+            to_json(j, options);
```
RMeli:
```suggestion
            nlohmann::json j = options;
```

metatomic-core/include/metatomic/system.hpp
Luthaf: Thanks, this seems good. I would rather keep `nlohmann::json` as an internal dependency, and not have users rely on it (so we can swap for something else in the future if needed).

PR #292, opened 2026-07-21
Python Build

johannes-spies:
This PR adds the infrastructure needed to build the `metatomic_core` Python wheel. Essentially, it's a port of the same functionality in `metatensor` ([`setup.py`,](https://github.com/metatensor/metatensor/blob/main/python/metatensor_core/setup.py) [`CMakeLists.txt`,](https://github.com/metatensor/metatensor/blob/main/python/metatensor_core/CMakeLists.txt) and [`package-core.sh`](https://github.com/metatensor/metatensor/blob/main/scripts/package-core.sh)). The only difference is that we import `metatensor` in `cmake_ext`'s `run` method to get the `cmake_prefix_path` which is required to build the native library.

The PR also makes an important change to the CI setup: now, containers for testing are built with Rust, similar to what `metatensor` does.

[📚 Download documentation for this pull-request](https://nightly.link/metatensor/metatomic/actions/artifacts/10103984340.zip)

[⚙️ Download Python wheels for this pull-request (you can install these with pip)](https://nightly.link/metatensor/metatomic/actions/artifacts/10104598844.zip)

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -24,6 +24,79 @@ concurrency:
 
 
 jobs:
+  build-core-wheels:
+    runs-on: ${{ matrix.os }}
+    name: ${{ matrix.name }}
+    strategy:
+      matrix:
+        include:
+          - name: x86_64 Linux
+            os: ubuntu-24.04
+            rust-target: x86_64-unknown-linux-gnu
+            cibw-arch: x86_64
+          - name: arm64 Linux
+            os: ubuntu-24.04-arm
+            rust-target: aarch64-unknown-linux-gnu
+            cibw-arch: aarch64
+          - name: arm64 macOS
+            os: macos-15
+            rust-target: aarch64-apple-darwin
+            cibw-arch: arm64
+          - name: x86_64 Windows
+            os: windows-2022
+            rust-target: x86_64-pc-windows-msvc
+            cibw-arch: AMD64
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+
+      - name: setup rust
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable
+          target: ${{ matrix.rust-target }}
+
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.14"
+
+      - name: install dependencies
+        run: python -m pip install cibuildwheel twine
+
+      - name: build custom manylinux docker images
+        if: matrix.os == 'ubuntu-24.04' || matrix.os == 'ubuntu-24.04-arm'
+        run: |
+          docker buildx build \
+              -t rustc-manylinux_2_28_${{ matrix.cibw-arch }} \
+              python/scripts/rustc-manylinux_2_28_${{ matrix.cibw-arch }}
+
+      - name: build metatomic-core wheel
+        run: python -m cibuildwheel python/metatomic_core
+        env:
+          CIBW_BUILD: cp311-*
+          CIBW_SKIP: "*musllinux*"
+          CIBW_ARCHS: ${{ matrix.cibw-arch }}
+          CIBW_BUILD_VERBOSITY: 1
+          CIBW_MANYLINUX_X86_64_IMAGE: rustc-manylinux_2_28_x86_64
+          CIBW_MANYLINUX_AARCH64_IMAGE: rustc-manylinux_2_28_aarch64
+          CIBW_ENVIRONMENT: >
+            MACOSX_DEPLOYMENT_TARGET=11
+          # do not complain for missing shared libraries provided by other packages
+          CIBW_REPAIR_WHEEL_COMMAND_MACOS: |
+            delocate-wheel --ignore-missing-dependencies --require-archs {delocate_archs} -w {dest_dir} -v {wheel}
+          CIBW_REPAIR_WHEEL_COMMAND_LINUX: |
+            auditwheel repair --exclude libmetatensor.so -w {dest_dir} {wheel}
+
+      - name: check wheels with twine
+        run: twine check wheelhouse/*.whl
+
+      - uses: actions/upload-artifact@v7
+        with:
+          name: core-wheel-${{ matrix.os }}-${{ matrix.cibw-arch }}
+          path: ./wheelhouse/*.whl
+
   build-torch-wheels:
     runs-on: ${{ matrix.os }}
     name: ${{ matrix.name }} (torch v${{ matrix.torch-version }})
@@ -89,8 +162,8 @@ jobs:
         if: matrix.os == 'ubuntu-24.04' || matrix.os == 'ubuntu-24.04-arm'
         run: |
           docker buildx build \
-              -t gcc11-manylinux_2_28_${{ matrix.cibw-arch }} \
-              python/scripts/gcc11-manylinux_2_28_${{ matrix.cibw-arch }}
+              -t rustc-manylinux_2_28_${{ matrix.cibw-arch }} \
+              python/scripts/rustc-manylinux_2_28_${{ matrix.cibw-arch }}
 
       - name: build metatomic-torch wheel
         run: python -m cibuildwheel python/metatomic_torch
@@ -99,8 +172,8 @@ jobs:
           CIBW_SKIP: "*musllinux*"
           CIBW_ARCHS: ${{ matrix.cibw-arch }}
           CIBW_BUILD_VERBOSITY: 1
-          CIBW_MANYLINUX_X86_64_IMAGE: gcc11-manylinux_2_28_x86_64
-          CIBW_MANYLINUX_AARCH64_IMAGE: gcc11-manylinux_2_28_aarch64
+          CIBW_MANYLINUX_X86_64_IMAGE: rustc-manylinux_2_28_x86_64
+          CIBW_MANYLINUX_AARCH64_IMAGE: rustc-manylinux_2_28_aarch64
           # METATOMIC_NO_LOCAL_DEPS is set to 1 when building a tag of
           # metatomic-torch, which will force to use the version of
           # metatomic-core already released on PyPI. Otherwise, this will use
@@ -214,6 +287,9 @@ jobs:
       - name: install dependencies
         run: python -m pip install wheel build twine
 
+      - name: build metatomic-core sdist
+        run: python -m build python/metatomic_core --sdist --outdir=dist/
+
       - name: build metatomic-torch sdist
         run: python -m build python/metatomic_torch --sdist --outdir=dist/
 
@@ -231,6 +307,7 @@ jobs:
 
       - name: create C++ tarballs
         run: |
+          ./scripts/package-core.sh dist/cxx/
           ./scripts/package-torch.sh dist/cxx/
 
       - uses: actions/upload-artifact@v7
@@ -243,12 +320,19 @@ jobs:
 
   merge-and-release:
     name: Merge and release wheels/sdists
-    needs: [merge-torch-wheels, build-others]
+    needs: [build-core-wheels, merge-torch-wheels, build-others]
     runs-on: ubuntu-24.04
     permissions:
       contents: write
       pull-requests: write
     steps:
+      - name: Download metatomic-core wheels
+        uses: actions/download-artifact@v8
+        with:
+          path: wheels
+          pattern: core-wheel-*
+          merge-multiple: true
+
       - name: Download metatomic-torch wheels
         uses: actions/download-artifact@v8
         with:
@@ -276,6 +360,17 @@ jobs:
         with:
           name: wheels
           description: ⚙️ Download Python wheels for this pull-request (you can install these with pip)
+
+      - name: upload to GitHub release (metatomic-core)
+        if: startsWith(github.ref, 'refs/tags/metatomic-core-v')
+        uses: softprops/action-gh-release@v3
+        with:
+          files: |
+            wheels/cxx/metatomic-core-cxx-*.tar.gz
+            wheels/metatomic_core-*
+          prerelease: ${{ contains(github.ref, '-rc') }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: upload to GitHub release (metatomic)
         if: startsWith(github.ref, 'refs/tags/metatomic-v')
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -18,35 +18,26 @@ jobs:
       matrix:
         include:
           - os: ubuntu-24.04
-            python-version: "3.10"
+            python-version: "3.11"
             torch-version: "2.3"
             numpy-version-pin: "<2.0"
-            # Do not run docs-tests with python 3.10 since torch-sim-atomistic
-            # is not available for this version of python
-            tox-envs: lint,torch-tests
           - os: ubuntu-24.04
-            python-version: "3.10"
+            python-version: "3.11"
             torch-version: "2.13"
-            # See above
-            tox-envs: lint,torch-tests
           - os: ubuntu-24.04
             # TorchScript is no longer supported in Python 3.14
             # so we keep a test with 3.13 to make sure this doesn't break
             python-version: "3.13"
             torch-version: "2.13"
-            tox-envs: lint,torch-tests,docs-tests
           - os: ubuntu-24.04
             python-version: "3.14"
             torch-version: "2.13"
-            tox-envs: lint,torch-tests,docs-tests
           - os: macos-15
             python-version: "3.14"
             torch-version: "2.13"
-            tox-envs: lint,torch-tests,docs-tests
           - os: windows-2022
             python-version: "3.14"
             torch-version: "2.13"
-            tox-envs: lint,torch-tests,docs-tests
     steps:
       - uses: actions/checkout@v7
         with:
@@ -85,7 +76,7 @@ jobs:
           python -m pip install tox coverage
 
       - name: run tests
-        run: tox -e ${{ matrix.tox-envs }}
+        run: tox -e lint,torch-tests,docs-tests
         env:
           PIP_EXTRA_INDEX_URL: https://download.pytorch.org/whl/cpu
           METATOMIC_TESTS_TORCH_VERSION: ${{ matrix.torch-version }}
--- a/.github/workflows/torch-tests.yml
+++ b/.github/workflows/torch-tests.yml
@@ -43,13 +43,16 @@ jobs:
     steps:
       - name: install dependencies in container
         if: matrix.container == 'ubuntu:22.04'
+        env:
+          # python3.11 pulls in tzdata, which otherwise asks for a timezone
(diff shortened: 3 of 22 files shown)


.github/workflows/build-wheels.yml
```diff
@@ -77,20 +77,36 @@ jobs:
...
 
-      - name: Set up Python
-        uses: actions/setup-python@v6
+      - name: setup rust
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable
+          target: ${{ matrix.rust-target }}
+
+      - name: Install uv
```
Luthaf: let's not use uv here, it is not giving us much

.github/workflows/torch-tests.yml
```diff
@@ -70,6 +70,13 @@ jobs:
         with:
           toolchain: stable
 
+      # the pip installs done by the tests below prefer `uv` when available,
+      # following metatensor's setup
+      - name: Install uv
+        uses: astral-sh/setup-uv@v8.2.0
+        with:
+          enable-cache: true
+
```
Luthaf:
```suggestion
```

.github/workflows/torch-tests.yml
```diff
@@ -94,6 +101,7 @@ jobs:
         env:
           # Use the CPU only version of torch when building/running the code
           PIP_EXTRA_INDEX_URL: https://download.pytorch.org/whl/cpu
+          UV_EXTRA_INDEX_URL: https://download.pytorch.org/whl/cpu
```
Luthaf:
```suggestion
```

python/metatomic_ase/setup.py
```diff
@@ -77,12 +77,16 @@ def git_version_info():
...
                 f"stderr: {output.stderr}\n"
             )
-        elif output.stderr:
+        if output.stderr:
             print(output.stderr, file=sys.stderr)
+
+        lines = output.stdout.splitlines()
+        if len(lines) < 2:
+            # the script gave up early (cf. `warn_and_exit`), and only printed
+            # the number of commits
```
Luthaf:
```suggestion
            # the script gave up early (cf. `warn_and_exit`)
```

python/metatomic_core/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+
+cmake_minimum_required(VERSION 3.22)
+project(metatomic-python NONE)
+
+option(METATOMIC_CORE_PYTHON_USE_EXTERNAL_LIB "Force the usage of an external version of metatomic-core" OFF)
+set(METATOMIC_CORE_SOURCE_DIR "" CACHE PATH "Path to the sources of metatomic-core")
+
+file(REMOVE ${CMAKE_INSTALL_PREFIX}/_external.py)
+
+set(REQUIRED_METATOMIC_VERSION "0.2.3")
```
Luthaf: err, this version does not exist =)

python/metatomic_torch/setup.py
```diff
@@ -200,12 +200,16 @@ def git_version_info():
...
                 f"stderr: {output.stderr}\n"
             )
-        elif output.stderr:
+        if output.stderr:
             print(output.stderr, file=sys.stderr)
+
+        lines = output.stdout.splitlines()
+        if len(lines) < 2:
+            # the script gave up early (cf. `warn_and_exit`), and only printed
+            # the number of commits
```
Luthaf:
```suggestion
            # the script gave up early (cf. `warn_and_exit`)
```

python/metatomic_torchsim/setup.py
```diff
@@ -62,12 +62,16 @@ def git_version_info():
...
                 f"stderr: {output.stderr}\n"
             )
-        elif output.stderr:
+        if output.stderr:
             print(output.stderr, file=sys.stderr)
+
+        lines = output.stdout.splitlines()
+        if len(lines) < 2:
+            # the script gave up early (cf. `warn_and_exit`), and only printed
+            # the number of commits
```
Luthaf:
```suggestion
        if len(lines) < 2:
            # the script gave up early (cf. `warn_and_exit`)
```

python/scripts/gcc11-manylinux_2_28_aarch64/Dockerfile
Luthaf, line 1: don't we still need this one for building metatomic-torch wheels?

scripts/git-version-info.py
```diff
@@ -174,7 +174,14 @@ def git_hash_all_code():
...
             git_env["GIT_INDEX_FILE"] = tmp.name
-            run_subprocess(["git", "add", "--all"], env=git_env, cwd=worktree)
+            # `core.safecrlf=false` since we are only staging files to compute a hash,
+            # and never write them back to the working tree: the end-of-line warnings
+            # git would emit here (on Windows) are only noise.
+            run_subprocess(
+                ["git", "-c", "core.safecrlf=false", "add", "--all"],
+                env=git_env,
+                cwd=worktree,
+            )
```
Luthaf, line 184: This would be nice to also port to metatensor

Luthaf (commented):
Looks good overall, I'm not convinced we should switch to uv in CI, and we should for sure do it in a separate PR anyway

johannes-spies:
Thank you for the review! You're right, the switch to `uv` was not necessary and I will revert it. I just tried to make the CI as similar as possible to track down the failure in the windows tests.

.github/workflows/torch-tests.yml
johannes-spies: [8050921](https://github.com/metatensor/metatomic/pull/292/commits/805092146b78252227f5514e89dd9f9be2e51d00)

.github/workflows/torch-tests.yml
johannes-spies: [8050921](https://github.com/metatensor/metatomic/pull/292/commits/805092146b78252227f5514e89dd9f9be2e51d00)

.github/workflows/build-wheels.yml
johannes-spies: [8050921](https://github.com/metatensor/metatomic/pull/292/commits/805092146b78252227f5514e89dd9f9be2e51d00)

scripts/git-version-info.py
johannes-spies: https://github.com/metatensor/metatensor/pull/1167

python/metatomic_core/CMakeLists.txt
johannes-spies: Good catch. This was a leftover from copying the file from `metatensor`. Fixed in [17d495c](https://github.com/metatensor/metatomic/pull/292/commits/17d495cc77257954d7da85098b4ff634642dce99).

python/scripts/gcc11-manylinux_2_28_aarch64/Dockerfile
johannes-spies: No, I switched to the `rustc-manylinux` container as it also comes with an appropriate toolchain in the `.github/workflows/build-wheels.yml` file.

.github/workflows/build-wheels.yml
```diff
@@ -77,6 +77,12 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: setup rust
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable
+          target: ${{ matrix.rust-target }}
+
```
Luthaf: Should not be required, Rust is only used inside the container

.github/workflows/build-wheels.yml
```diff
@@ -85,12 +91,20 @@ jobs:
...
         run: python -m pip install cibuildwheel
 
-      - name: build custom manylinux docker images
+      - name: Set up Docker Buildx
         if: matrix.os == 'ubuntu-24.04' || matrix.os == 'ubuntu-24.04-arm'
-        run: |
-          docker buildx build \
-              -t gcc11-manylinux_2_28_${{ matrix.cibw-arch }} \
-              python/scripts/gcc11-manylinux_2_28_${{ matrix.cibw-arch }}
+        uses: docker/setup-buildx-action@v4
```
Luthaf: Why is this needed now when it worked without it before?

.github/workflows/build-wheels.yml
```diff
@@ -85,12 +91,20 @@ jobs:
...
+
+      - name: build manylinux with rust docker image
+        if: matrix.os == 'ubuntu-24.04' || matrix.os == 'ubuntu-24.04-arm'
+        uses: docker/build-push-action@v7
```
Luthaf: Why use a custom action?

Luthaf (commented):
The wheel (as created by `python -m build python/metatomic_core`) contains configuration files to find `nlohmann::json`, which I don't think we want to make available through the to external users here.

```
[...]
adding 'metatomic/share/cmake/nlohmann_json/nlohmann_jsonConfig.cmake'
adding 'metatomic/share/cmake/nlohmann_json/nlohmann_jsonConfigVersion.cmake'
adding 'metatomic/share/cmake/nlohmann_json/nlohmann_jsonTargets.cmake'
adding 'metatomic/share/pkgconfig/nlohmann_json.pc'
[...]
```

Ideally they should not even be installed by the main metatomic CMakeLists.txt (this is not specific to Python) when we fetch the code from github ourself.

Can you have a look?

johannes-spies:
Yes, you're right. [27675fd](https://github.com/metatensor/metatomic/pull/292/commits/27675fde208f46fb1c052e42d55b958a462a61f0) sets `JSON_Install` to `OFF`.

Luthaf:
We still need to install the headers, just not the cmake config / `.pc` files

.github/workflows/build-wheels.yml
johannes-spies: [037a124](https://github.com/metatensor/metatomic/pull/292/commits/037a1244c72b852045b421be7c966b4475c33572)

.github/workflows/build-wheels.yml
johannes-spies: c1498bd5

.github/workflows/build-wheels.yml
johannes-spies: 5f151c64

johannes-spies:
I added the headers in 8dc3972a. Also, I tried making the `build-wheels.yml` pipelines as close as possible to metatensor's pipeline! I hope everything is now very similar, minus `uv`.

metatomic-core/CMakeLists.txt
```diff
@@ -461,3 +461,14 @@ install(FILES
...
+if (NOT nlohmann_json_FOUND)
+    # nlohmann_json was fetched and vendored by us: install its headers (used
+    # by metatomic's own public headers), but remove the CMake package config
+    # and pkg-config files it installs, since we don't want to advertise a
+    # system-wide nlohmann_json package to external users.
+    install(CODE "
+        file(REMOVE_RECURSE \"\${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_DATADIR}/cmake/nlohmann_json\")
+        file(REMOVE \"\${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_DATADIR}/pkgconfig/nlohmann_json.pc\")
+    ")
+endif()
```
Luthaf, line 474: Sorry for the back and forth here, I did not realized we were using the cmake files from our own metatomic-config.cmake …

Maybe the best solution here would be to install these to a different directory, i.e. to `CMAKE_INSTALL_PREFIX/lib/metatomic/external` instead of `CMAKE_INSTALL_PREFIX`? This way we can find the headers with normal cmake configuration, and also hide them from most users by default.

@RMeli what do you think here?

scripts/git-version-info.py
Luthaf: Can you also send this as a standalone commit to main on this repo?

python/metatomic_core/setup.py
```diff
@@ -47,18 +150,66 @@ def run(self):
...
+    if output.returncode != 0:
+        stderr = output.stderr
+        stdout = output.stdout
+        raise RuntimeError(
+            "failed to collect C++ sources for Python sdist\n"
+            f"stdout:\n {stdout}\n\nstderr:\n {stderr}"
+        )
+
+
+def get_rust_version():
```
RMeli, line 193: I'd use `tomllib` instead of manual parsing (which seems a bit fragile).

python/metatomic_core/setup.py
```diff
@@ -1,25 +1,128 @@
...
+
+class cmake_ext(build_ext):
+    """
+    Build the native library using cmake
+    """
+
+    def finalize_options(self):
+        if self.editable_mode:
+            raise RuntimeError(
+                "metatensor-core does not support editable installation yet"
```
RMeli:
```suggestion
                "metatomic-core does not support editable installation yet"
```
?

scripts/package-core.sh
```diff
@@ -0,0 +1,54 @@
...
+cp "$ROOT_DIR/README.md" "$TMP_DIR/$ARCHIVE_NAME"
+
+# Get the git version information, this is used when building the
+# code to change the version for development builds
+cd "$ROOT_DIR"
+./scripts/git-version-info.py "metatomic-core-v" > "$TMP_DIR/$ARCHIVE_NAME/cmake/git_version_info"
+
+cd "$TMP_DIR"
+# Compile metatomic-core as it's own Cargo workspace (otherwise we can not the
+# use metatomic rust crate in a project using workspaces).
```
RMeli:
```suggestion
# Compile metatomic-core as it's own Cargo workspace (otherwise we can not use
# the metatomic rust crate in a project using workspaces).
```

python/scripts/rustc-manylinux_2_28_x86_64/Dockerfile
```diff
@@ -0,0 +1,18 @@
+# Use manylinux docker image as a base
+FROM quay.io/pypa/manylinux_2_28_x86_64
+
+RUN yum install git -y
+RUN git config --global --add safe.directory /code
+
+# Download rustup-init asn install
```
RMeli:
```suggestion
# Download rustup-init and install
```

metatomic-core/CMakeLists.txt
RMeli: Sorry, I totally missed this ping! Hiding the installation sounds like a good idea. Another alternative is to simply manually install the headers, given that the library is header-only (there is also a single-header option).

python/scripts/rustc-manylinux_2_28_aarch64/Dockerfile
```diff
@@ -0,0 +1,18 @@
+# Use manylinux docker image as a base
+FROM quay.io/pypa/manylinux_2_28_aarch64
+
+RUN yum install git -y
+RUN git config --global --add safe.directory /code
+
+# Download rustup-init asn install
```
RMeli:
```suggestion
# Download rustup-init and install
```

python/scripts/rustc-manylinux_2_28_aarch64/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --profile minimal --default-toolchain 1.96
```
RMeli: Worth making the default toolchain an `ARG` (with default value `1.96`)?

python/scripts/rustc-manylinux_2_28_x86_64/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --profile minimal --default-toolchain 1.96
```
RMeli: Same as above.

python/metatomic_core/setup.py
johannes-spies: Good idea, but `tomllib` was added in Python 3.11 and [the minimum python version is set to 3.10](https://github.com/metatensor/metatomic/blob/1637cfb5a69a10c04bc1c29c3ee1f48d60c326f0/pyproject.toml#L5) in this project. I am not sure if we want to bump it because of `metatomic-core`'s `setup.py`. I don't mind bumping it, so should we? @Luthaf, do you have an opinion on this?

Luthaf: Python 3.10 is going to be EOL in 2 months, so likely around the same time we release this. So I'm not against bumping the minimal version to 3.11 already!

metatomic-core/CMakeLists.txt
johannes-spies: I tried to include it header-only in [25d8f18](https://github.com/metatensor/metatomic/pull/292/commits/25d8f18df985e26e99310c3ea2f36d47d82c5331).

python/metatomic_core/setup.py
johannes-spies: [522fdaa](https://github.com/metatensor/metatomic/pull/292/commits/522fdaad7ab722e5fd271c2c4f630c42ccad8919)

python/scripts/rustc-manylinux_2_28_x86_64/Dockerfile
johannes-spies: [69b2af1](https://github.com/metatensor/metatomic/pull/292/commits/69b2af1dfa78ad8b759954a6f3d55a9b48cff12e)

python/scripts/rustc-manylinux_2_28_aarch64/Dockerfile
johannes-spies: [69b2af1](https://github.com/metatensor/metatomic/pull/292/commits/69b2af1dfa78ad8b759954a6f3d55a9b48cff12e)

.github/workflows/python-tests.yml
```diff
@@ -18,14 +18,14 @@ jobs:
       matrix:
         include:
           - os: ubuntu-24.04
-            python-version: "3.10"
+            python-version: "3.11"
             torch-version: "2.3"
             numpy-version-pin: "<2.0"
-            # Do not run docs-tests with python 3.10 since torch-sim-atomistic
+            # Do not run docs-tests with python 3.11 since torch-sim-atomistic
```
Luthaf: This might be wrong now, so we can now run doc-tests with python 3.11. I'll let you check though!

metatomic-core/cmake/nlohmann_json.cmake
```diff
@@ -31,10 +31,15 @@ else()
...
-  # config / pkg-config files, which we strip after install below.
-  set(JSON_Install ON CACHE INTERNAL "")
+  # Don't use nlohmann_json's own install rules, they would also install its
+  # CMake package config and pkg-config files, and we don't want to advertise
+  # a system-wide nlohmann_json package to external users.
+  set(JSON_Install OFF CACHE INTERNAL "")
 
   FetchContent_MakeAvailable(nlohmann_json)
+
+  # nlohmann_json is header-only, so we can install the headers ourselves,
```
Luthaf, line 41: so the cmake config file handles both the headers being installed by us and us re-using a system install of nlohmann_json, right?

.github/workflows/python-tests.yml
Luthaf: running on CI rn!

Luthaf (approved):

Issue #271, opened 2026-06-26
Add a way to query the components/properties associated with an output ahead of time


Luthaf:
In multiple cases (PLUMED integration, lammps' `compute metatomic`) it is required to know the shape of the output **before** we even can call the model. In PLUMED we end up calling the model with 0 atoms just to extract this information, but in general it would be useful to have an API for this on the models, maybe something like 

```
class MyModel:

    def output_layout(self, outputs: Dict[str, ModelOutput]) -> Dict[str, TensorMap]:
        ...
```

With the function returning 0-samples TensorMap with the correct metadata. We could even provide a function to implement this easily for outputs with statically known shapes (i.e. most standard outputs)

GardevoirX:
See https://github.com/metatensor/lammps/pull/46#discussion_r3550653158, we can add also the expected metadata of the standard quantities here

PR #294, opened 2026-07-23
Add an O(3)-symmetrized atomistic model wrapper

MichelangeloDomina:
## Summary

Adds `metatomic.torch.symmetrized_model`: wrap an exported `AtomisticModel` to
average its predictions over O(3) with a finite rotation quadrature, and to
measure how far the underlying model is from being equivariant.

Everything is selected through regular output requests:

| Requested output | Result |
| --- | --- |
| `<name>` | O(3) average of the underlying output |
| `o3::variance::<name>` | per-sample equivariance variance |
| `o3::character_projection::<name>` | squared character-projection contribution per irrep |

`SymmetrizedModel.wrap()` returns a regular exportable `AtomisticModel`
(metadata, capabilities, neighbor lists, and custom inputs are carried over),
so the wrapped model can be used anywhere the original could, including
TorchScript export.

The conventions (quadrature, variance definition, irrep bookkeeping in
TensorMap keys) are documented in `docs/src/torch/reference/symmetrized-model.rst`.

## Limitations

- the quadrature is exact only up to a maximal angular momentum; convergence
  should be checked by increasing `max_o3_lambda_grid`;
- CPU and CUDA only, float32/float64;
- explicit TensorBlock gradients are not supported (PyTorch autograd through
  returned values works).

- ~~[ ] Issue referenced (for PRs that solve an issue)?~~

# Reviewer checklist


docs/src/torch/reference/symmetrized-model.rst
```diff
@@ -0,0 +1,144 @@
+.. _symmetrized-model:
+
+O(3)-symmetrized models
+=======================
+
+The :py:mod:`metatomic.torch.symmetrized_model` module wraps an exported
```
Luthaf: Let's export it as `metatomic.torch.SymmetrizedModel`, and make the module private

Luthaf:
```suggestion
The :py:mod:`metatomic.torch.symmetrized_model` module wraps an existing
```

docs/src/torch/reference/symmetrized-model.rst
```diff
@@ -0,0 +1,144 @@
...
+:py:class:`~metatomic.torch.AtomisticModel` with finite-quadrature O(3)
+averaging and equivariance diagnostics. Ordinary outputs are averaged over
+rotated and inverted copies of each input. Additional output names request an
+equivariance variance or squared character-projection contributions of the
+model response.
```
Luthaf:
```suggestion
averaging and equivariance diagnostics. Pre-existing outputs of the model are
averaged over rotated and inverted copies of each input.
:py:class:`SymmetrizedModel` also add extra outputs to compute the equivariance
variance or squared character-projection contributions of the model response.
```

docs/src/torch/reference/symmetrized-model.rst
```diff
@@ -0,0 +1,144 @@
...
+     - O(3) average of the underlying ``<name>`` output
+   * - ``o3::variance::<name>``
+     - component-averaged equivariance variance of ``<name>``
+   * - ``o3::character_projection::<name>``
+     - unnormalized squared character-projection contributions of ``<name>``
+
+``<name>`` is preserved verbatim. It can therefore be a standard quantity, a
+variant such as ``energy/pbe``, or a custom name such as
+``mtt::feature::node``. For example,
+``o3::variance::energy/pbe`` evaluates the underlying ``energy/pbe`` output.
```
Luthaf, line 34: This is not very clear, but I can rephrase when everything else is done here =)

docs/src/torch/reference/symmetrized-model.rst
```diff
@@ -0,0 +1,144 @@
...
+
+Reference
+---------
+
+.. py:currentmodule:: metatomic.torch.symmetrized_model
+
+.. autoclass:: SymmetrizedModel
+    :members:
+
+.. autofunction:: get_rotation_quadrature
```
Luthaf: why is this part of the public API?

python/metatomic_torch/metatomic/torch/symmetrized_model/__init__.py
Luthaf, line 1: can we rename the module to something like  `_o3_symmetrized`? Or maybe we can put this in the existing `_o3` module?

python/metatomic_torch/metatomic/torch/o3/_decompose.py
Luthaf, line 1: could you add a quick docstring to all modules explaining what they do?

python/metatomic_torch/metatomic/torch/o3/_decompose.py
```diff
@@ -0,0 +1,211 @@
...
+
+def _cartesian_vectors_to_spherical(
+    values: torch.Tensor,
+    component_axis: int,
+) -> torch.Tensor:
+    """Reorder ``(x, y, z)`` as ``(mu=-1, 0, 1) = (y, z, x)``."""
+    return values.roll(-1, dims=component_axis)
+
+
+def _symmetric_matrices_to_spherical(
```
Luthaf: does this have requirements on the input matrix shape?

python/metatomic_torch/metatomic/torch/symmetrized_model/_decompose.py
```diff
@@ -0,0 +1,211 @@
...
+    )
+
+    return l0, l2
+
+
+def _decompose_output(
+    source_name: str,
+    tensor: TensorMap,
+) -> TensorMap:
+    """Decompose standard outputs for variance and character projection."""
```
Luthaf: so "decompose" means take existing standard outputs on a cartesian/scalar basis and express them on the usual o3 convention?

python/metatomic_torch/metatomic/torch/symmetrized_model/_decompose.py
```diff
@@ -0,0 +1,211 @@
...
+    quantity = source_name.split("/", 1)[0]
+    is_energy = quantity in (
```
Luthaf: maybe this should handle other scalars?

python/metatomic_torch/metatomic/torch/symmetrized_model/_decompose.py
```diff
@@ -0,0 +1,211 @@
...
+        "energy",
+        "energy_ensemble",
+        "energy_uncertainty",
+    )
+    is_force = quantity in (
+        "non_conservative_force",
+        "non_conservative_forces",
+    )
```
Luthaf: The code should not handle old-style names, only the new ones

python/metatomic_torch/metatomic/torch/symmetrized_model/_decompose.py
```diff
@@ -0,0 +1,211 @@
...
+                    components=[_o3_mu_labels(0, block.values.device)],
+                    properties=block.properties,
+                )
+            )
+        result = TensorMap(
+            _add_o3_irrep_to_keys(tensor.keys, 0, 1),
+            energy_blocks,
+        )
+
+    elif is_force:
```
Luthaf: should this deal with other "vector" quantities? Such as position/velocity/momentum?

python/metatomic_torch/metatomic/torch/o3/_decompose.py
```diff
@@ -0,0 +1,211 @@
...
+                len(block.components) != 2
+                or block.components[0].names != ["xyz_1"]
+                or block.components[1].names != ["xyz_2"]
+                or len(block.components[0]) != 3
+                or len(block.components[1]) != 3
+            ):
+                raise ValueError(
+                    "non_conservative_stress must have 'xyz_1' and 'xyz_2' "
+                    "component axes of size 3"
+                )
```
Luthaf: This is already be checked elsewhere, the code here should assume it or use an assert

Luthaf: (same for the forces)

python/metatomic_torch/metatomic/torch/o3/_projections.py
```diff
@@ -0,0 +1,192 @@
...
+The projected quantity is defined in the :py:class:`SymmetrizedModel` class
+docstring.
+"""
+
+from typing import List, Tuple
+
+import torch
+from metatensor.torch import Labels, TensorBlock, TensorMap
+
+from ._utils import (
```
Luthaf, line 14: functions in ._utils (or this file) that are intended to be imported by other files should not start by an underscore.

Function in a class starting with underscore => the function is private to the class/the module that defines the class

Function in a module starting with underscore => the function is private to the module and should not be imported outside of it.

The module itself is private and should not be used by code outside of metatomic, but it can still provide public functions to be used outside of it.

This way, we easily see if a function is just a current implementation detail or something used outside of the current module/file.

`_quadrature.py` does this well

python/metatomic_torch/metatomic/torch/symmetrized_model/_projections.py
```diff
@@ -0,0 +1,192 @@
...
+            )
+
+    if len(output_key_values) == 0:
+        values = key_values.new_empty((0, len(key_names) + 1))
+    else:
+        values = torch.stack(output_key_values)
+    return TensorMap(Labels(key_names + ["chi_lambda"], values), blocks)
+
+
+def _character_projection_tensormap_from_cosets(
```
Luthaf: Could you explain to me what a coset is?

python/metatomic_torch/metatomic/torch/symmetrized_model/_wigner_storage.py
```diff
@@ -0,0 +1,59 @@
+import torch
+
+from ..o3 import O3Transformation
+
+
+def _build_packed_wigner_matrices(
```
Luthaf: I don't love that we have a different representation here and in `O3Transformation`.  This is just putting all matrices in the same tensor, right? I assume this is for efficiency purposes?

Could we change the format used by `O3Transformation` to be the same as this one?

python/metatomic_torch/metatomic/torch/symmetrized_model/_model.py
```diff
@@ -0,0 +1,1162 @@
...
+        self.register_buffer("_rotation_weights", rotation_weights)
+        self.register_buffer("_packed_wigner_matrices", packed_wigner_matrices)
+
+    @staticmethod
+    def wrap(
+        model: AtomisticModel,
+        *,
+        max_o3_lambda_target: int,
+        max_o3_lambda_input: int = 0,
+        max_o3_lambda_character: Optional[int] = None,
```
Luthaf: Can we auto-guess these from the model by default, by checking the capabilities and extra inputs against the list of standard quantities? And then allow to override for custom inputs/outputs?

python/metatomic_torch/metatomic/torch/symmetrized_model/_model.py
```diff
@@ -0,0 +1,1162 @@
...
+        variance_first_moments = torch.jit.annotate(Dict[str, TensorMap], {})
+        variance_second_moments = torch.jit.annotate(Dict[str, TensorMap], {})
+        variance_absolute_second_moments = torch.jit.annotate(
+            Dict[str, TensorMap],
+            {},
+        )
+        proper_character_coefficients = torch.jit.annotate(
+            Dict[str, TensorMap],
+            {},
+        )
```
Luthaf: I think this should work written as 
```suggestion
        proper_character_coefficients: Dict[str, TensorMap] = {}
```

Which should be easier to read

Luthaf (commented):
I only checked the code, I'll try to do a pass on the tests ASAP

python/metatomic_torch/metatomic/torch/symmetrized_model/_decompose.py
ppegolo: we can add them, but they're not really things PET can produce, but inputs (mass, charge, ...), right?

python/metatomic_torch/metatomic/torch/symmetrized_model/_decompose.py
ppegolo: This one yes

python/metatomic_torch/metatomic/torch/symmetrized_model/_projections.py
ppegolo: They are in general subsets of groups that can be used to split a group in some interesting way. In this case into proper (SO(3)) and improper (SO(3) * inversion) rotations. Coset is the correct word here (subgroup would be wrong and apply to the SO(3) case only). We could call it "subset", but we would be using a generic word instead of the specific one. I can add a comment or update the docs to explain it

python/metatomic_torch/metatomic/torch/symmetrized_model/_decompose.py
ppegolo: actually it costs nothing to just check if they're standard outputs unconditionally in case they're going to be predicted one day, so I'll do it

python/metatomic_torch/metatomic/torch/symmetrized_model/_projections.py
Luthaf: cosets is fine, just might be worth a small comment for people like me who forgot all their group theory

python/metatomic_torch/metatomic/torch/symmetrized_model/_model.py
Luthaf: For non-standard outputs/more complex things down the line, this would be another use case for https://github.com/metatensor/metatomic/issues/271

metatomic-torch/CHANGELOG.md
```diff
@@ -16,6 +16,12 @@ a changelog](https://keepachangelog.com/en/1.1.0/) format. This project follows
 ### Removed
 -->
 
+### Added
+
+- Added `metatomic.torch.SymmetrizedModel` for finite-quadrature O(3)
+  averaging, equivariance variances, and character projections of exported
```
Luthaf:
```suggestion
  averaging, equivariance variances, and character projections of existing
```

python/metatomic_torch/metatomic/torch/_quantities.py
```diff
@@ -0,0 +1,79 @@
+"""
+Python-side mirror of ``metatomic-torch/src/quantities.cpp`` (``KNOWN_QUANTITIES``
+and the per-quantity checks), holding the metadata for standard quantities: their
+category (Cartesian layout and spherical character) and their deprecated-name
+aliases.
+
+This module must not import anything from ``metatomic``, so that any other module
+can import it without creating an import cycle.
```
Luthaf:
```suggestion
Cartesian layout and spherical character for standard quantities
```

python/metatomic_torch/metatomic/torch/model.py
```diff
@@ -395,26 +396,10 @@ def __init__(
...
-        self._new_names = {
-            "features": "feature",
-            "non_conservative_forces": "non_conservative_force",
-            "positions": "position",
-            "momenta": "momentum",
-            "masses": "mass",
-            "velocities": "velocity",
-            "charges": "charge",
-        }
+        self._new_names = dict(NEW_QUANTITY_NAMES)
```
Luthaf: why `dict()`? To make a copy?

python/metatomic_torch/metatomic/torch/o3/_decompose.py
```diff
@@ -0,0 +1,218 @@
...
+    """Reorder ``(x, y, z)`` as ``(mu=-1, 0, 1) = (y, z, x)``."""
+    return values.roll(-1, dims=component_axis)
+
+
+def _symmetric_matrices_to_spherical(
+    values: torch.Tensor,
+) -> tuple[torch.Tensor, torch.Tensor]:
+    """Return orthonormal l=0 and l=2 components of the symmetric matrix part.
+
+    ``values`` must have shape ``(n_samples, 3, 3, n_properties)``.
```
Luthaf: Can we have an assert for this?

python/metatomic_torch/metatomic/torch/o3/_decompose.py
```diff
@@ -0,0 +1,218 @@
...
+
+    The antisymmetric (l=1) part is silently discarded.
```
Luthaf: Is this because it is expected to be small/negligible?

python/metatomic_torch/metatomic/torch/o3/_decompose.py
```diff
@@ -0,0 +1,218 @@
...
+            (values[:, 0, 2, :] + values[:, 2, 0, :]) / sqrt_two,
+            (values[:, 0, 0, :] - values[:, 1, 1, :]) / sqrt_two,
+        ],
+        dim=1,
+    )
+
+    return l0, l2
+
+
+def decompose_output(
```
Luthaf:
```suggestion
def decompose_quantity(
```

python/metatomic_torch/metatomic/torch/o3/_decompose.py
```diff
@@ -0,0 +1,218 @@
...
+    source_name: str,
+    tensor: TensorMap,
+) -> TensorMap:
+    """Decompose standard outputs for variance and character projection.
```
Luthaf:
```suggestion
    """Decompose standard quantities for variance and character projection.
```

python/metatomic_torch/metatomic/torch/o3/_decompose.py
```diff
@@ -0,0 +1,218 @@
...
+
+    This takes the standard Cartesian or scalar outputs of a model and
```
Luthaf:
```suggestion
    This takes the standard Cartesian or scalar quantities (inputs or outputs of a model) and
```

python/metatomic_torch/metatomic/torch/o3/_decompose.py
```diff
@@ -0,0 +1,218 @@
...
+            (values[:, 0, 0, :] - values[:, 1, 1, :]) / sqrt_two,
+        ],
+        dim=1,
+    )
+
+    return l0, l2
+
+
+def decompose_output(
+    source_name: str,
```
Luthaf:
```suggestion
    name: str,
```

python/metatomic_torch/metatomic/torch/o3/_decompose.py
```diff
@@ -0,0 +1,218 @@
...
+                    components=[_o3_mu_labels(1, block.values.device)],
+                    properties=block.properties,
+                )
+            )
+        result = TensorMap(
+            _add_o3_irrep_to_keys(tensor.keys, 1, 1),
+            vector_blocks,
+        )
+
+    else:
```
Luthaf, line 153:
```suggestion
    else:
        assert(category == "symmetric_matrix")
```

python/metatomic_torch/metatomic/torch/o3/_symmetrized.py
```diff
@@ -0,0 +1,1265 @@
...
+)
+
+
+def _use_new_quantity_name(name: str, new_names: Dict[str, str]) -> str:
+    """Replace a deprecated base quantity in ``name`` with its current name."""
+    parts = name.split("/")
+    if parts[0] in new_names:
+        parts[0] = new_names[parts[0]]
+        return "/".join(parts)
+    return name
```
Luthaf: Not sure what this is for? Transforming the output names to the new names even if the underlying model is using the deprecated ones?

python/metatomic_torch/metatomic/torch/o3/_symmetrized.py
```diff
@@ -0,0 +1,1265 @@
...
+
+
+def _transform_system_geometry_batch(
```
Luthaf: why is this not using O3Transformation?

python/metatomic_torch/metatomic/torch/o3/_symmetrized.py
```diff
@@ -0,0 +1,1265 @@
...
+    """Check a TensorMap's spherical component ranks against one limit."""
+    tensor_max_o3_lambda = _max_o3_lambda_in_tensor(tensor)
+    if tensor_max_o3_lambda > max_angular_momentum:
+        raise ValueError(
+            f"{tensor_description} contains o3_lambda={tensor_max_o3_lambda}, "
+            f"exceeding {limit_name}={max_angular_momentum}"
+        )
+
+
+def _transform_system_batch(
```
Luthaf: Same, this should use the existing code in O3Transformation instead of having a separate copy here

python/metatomic_torch/metatomic/torch/o3/_symmetrized.py
```diff
@@ -0,0 +1,1265 @@
...
+    if len(source_name) == 0:
+        raise ValueError(
+            f"requested output '{requested_name}' does not identify an "
+            "underlying model output"
+        )
+
+    return source_name, calculation
+
+
+def _record_output_request(
```
Luthaf: I don't understand this function, what is it checking for?

python/metatomic_torch/metatomic/torch/o3/_symmetrized.py
```diff
@@ -0,0 +1,1265 @@
...
+    reference: Optional[TensorMap],
+    compute_second_moments: bool,
+) -> Tuple[
+    TensorMap,
+    Optional[TensorMap],
+    Optional[TensorMap],
+    TensorMap,
+]:
+    """Accumulate one rotation batch's weighted moments, centered on a reference
+    value so the variance subtraction stays cancellation-safe."""
```
Luthaf: what does cancellation-safe means here? Can you extend the docstring?

python/metatomic_torch/metatomic/torch/o3/_symmetrized.py
```diff
@@ -0,0 +1,1265 @@
...
+
+def _clamp_roundoff_negative_diagnostic(
+    tensor: TensorMap,
+    scale: TensorMap,
+    *,
+    n_grid_points: int,
+    quantity: str,
+    max_angular_momentum_grid: int,
+) -> TensorMap:
+    """Clamp round-off negatives and reject invalid or materially negative values."""
```
Luthaf: Could the docstring explain why we are doing this?

python/metatomic_torch/metatomic/torch/o3/_symmetrized.py
```diff
@@ -0,0 +1,1265 @@
...
+
+        capabilities = model.capabilities()
+        supported_devices = [
+            device
+            for device in capabilities.supported_devices
+            if device == "cpu" or device == "cuda"
+        ]
+        if len(supported_devices) == 0:
+            raise ValueError(
+                "SymmetrizedModel supports CPU and CUDA execution, but the "
```
Luthaf: why limit the devices?

python/metatomic_torch/metatomic/torch/o3/_symmetrized.py
```diff
@@ -0,0 +1,1265 @@
...
+        if (
+            len(character_projection_names) != 0
+            and self.max_angular_momentum_character is None
+        ):
+            raise ValueError(
+                "max_angular_momentum_character must be set to request "
+                "character projections"
+            )
+
+        source_outputs = torch.jit.annotate(Dict[str, ModelOutput], {})
```
Luthaf: please use annotations instead of torch.jit.annotate

```suggestion
        source_outputs: Dict[str, ModelOutput] = {}
```

and same below

python/metatomic_torch/metatomic/torch/o3/_symmetrized.py
```diff
@@ -0,0 +1,1265 @@
...
+            for o3_lambda in range(self.max_angular_momentum_input + 1):
+                input_wigner_matrices.append(
+                    wigner_matrices_for_lambda(
+                        self._packed_wigner_matrices,
+                        n_rotations,
+                        o3_lambda,
+                    )[batch_start:batch_stop].to(
+                        dtype=work_dtype,
+                        device=work_device,
+                    )
```
Luthaf: Can we maybe use multiple lines and separate variables instead of trying to put eveything in one expression?

python/metatomic_torch/metatomic/torch/o3/_symmetrized.py
```diff
@@ -0,0 +1,1265 @@
...
+
+            inverse_target_wigner_matrices: List[torch.Tensor] = []
+            if needs_backrotation:
+                for o3_lambda in range(self.max_angular_momentum_target + 1):
+                    inverse_target_wigner_matrices.append(
+                        wigner_matrices_for_lambda(
+                            self._packed_wigner_matrices,
+                            n_rotations,
+                            o3_lambda,
+                        )[batch_start:batch_stop].transpose(1, 2)
```
Luthaf: Same

python/metatomic_torch/metatomic/torch/o3/_tranformations.py
```diff
@@ -206,30 +206,28 @@ def _create_no_checks(
...
     def _wigner_D_cache_entry(self, ell: int) -> torch.Tensor:
         """Return the internal cache entry for ``ell`` without copying it."""
         ell = self._validate_ell_range(ell)
 
-        D = self._ensure_wigner_D_cache().get(ell)
-        if D is None:
-            raise ValueError(f"Wigner-D matrix for ell={ell} not found in cache.")
-
-        return D
+        return wigner_matrices_for_lambda(self._ensure_wigner_D_cache(), 1, ell)[0]
```
Luthaf:
```suggestion
        return wigner_matrices_for_lambda(self._ensure_wigner_D_cache(), n_matrices=1, o3_lambda=ell)[0]
```

python/metatomic_torch/metatomic/torch/o3/_utils.py
```diff
@@ -0,0 +1,141 @@
...
+def validate_integer(name: str, value, minimum: int) -> int:
+    """Check that ``value`` is an integer at least ``minimum``.
+
+    Return it as a Python ``int``.
+    """
+    if isinstance(value, bool) or not isinstance(value, Integral):
+        raise TypeError(f"{name} must be an integer, got {type(value).__name__}")
+    integer_value = int(value)
+    if integer_value < minimum:
+        qualifier = "positive" if minimum == 1 else "non-negative"
```
Luthaf:
```suggestion
        if minimum == 0:
            qualifier = "non-negative"
         elif minimun == 1:
            qualifier = "positive" 
         else:
             qualifier = f"larger or equal to {minimum}"
```

Luthaf (commented):
Looks good overall, I mostly have small nits comments. 

The one larger thing is that we have two copies of the O3 transformation code, we should only have one.

I only looked at the code, second review for the test coming after that!

python/metatomic_torch/tests/symmetrized_model.py
```diff
@@ -0,0 +1,3101 @@
...
+                torch.empty((0, len(samples.names)), dtype=torch.int64),
+            ),
+            components=components,
+            properties=properties,
+        ),
+    )
+    return system
+
+
+class TestSystemGeometryBatch:
```
Luthaf: this should be tested in the other O3 tests for transformation

python/metatomic_torch/tests/symmetrized_model.py
```diff
@@ -0,0 +1,3101 @@
...
+
+        message = "system and matrices must have the same dtype and device"
+        with pytest.raises(ValueError, match=f"^{re.escape(message)}$"):
+            _transform_system_geometry_batch(
+                system,
+                torch.eye(3, dtype=torch.float32).unsqueeze(0),
+            )
+
+
+class TestSystemBatch:
```
Luthaf: Same

python/metatomic_torch/tests/symmetrized_model.py
```diff
@@ -0,0 +1,3101 @@
...
+            [0.50, -0.25, 0.30, 0.45],
+            dtype=torch.float32,
+        )
+        inverse_wigner_matrices = torch.randn(
+            (n_rotations, dimension, dimension),
+            dtype=torch.float32,
+            generator=generator,
+        )
+
+        coefficients = _character_projection_coefficients_from_rotation_batch(
```
Luthaf: we should not test private functions directly. Isn't this testable through the public API?

python/metatomic_torch/tests/symmetrized_model.py
```diff
@@ -0,0 +1,3101 @@
...
+                * inverse_wigner_matrices[rotation]
+                .to(torch.float64)
+                .reshape(1, dimension, dimension, 1, 1)
+                * values[rotation].reshape(n_samples, 1, 1, 2, 3)
+            )
+
+        assert torch.allclose(coefficients, expected, rtol=0.0, atol=1e-12)
+
+    @pytest.mark.parametrize("chi_lambda", [0, 1, 2])
+    def test_factorization_matches_all_rotation_pairs(self, chi_lambda):
```
Luthaf: same, private API

python/metatomic_torch/tests/symmetrized_model.py
```diff
@@ -0,0 +1,3101 @@
...
+                    )
+            expected.append(direct_sum)
+
+        assert torch.allclose(sigma_plus, expected[0], rtol=0.0, atol=1e-12)
+        assert torch.allclose(sigma_minus, expected[1], rtol=0.0, atol=1e-12)
+        assert torch.all(sigma_plus >= 0)
+        assert torch.all(sigma_minus >= 0)
+
+
+class TestWignerStorage:
```
Luthaf: this should not be needed once the code uses the existing O3 transformation code

python/metatomic_torch/tests/symmetrized_model.py
```diff
@@ -0,0 +1,3101 @@
...
+
+        assert len(o3_rotations) == 2 * len(rotations)
+        dets = np.linalg.det(o3_rotations)
+        assert np.allclose(np.sort(dets), np.repeat([-1.0, 1.0], len(rotations)))
+
+
+class TestSymmetrizedModelConstruction:
+    """Test construction of the quadrature and persistent Wigner-D storage."""
+
+    def test_constructs_registered_buffers(self):
```
Luthaf: not sure why this is worth testing

python/metatomic_torch/tests/symmetrized_model.py
```diff
@@ -0,0 +1,3101 @@
...
+        )
+        with pytest.raises(ValueError, match=f"^{re.escape(message)}$"):
+            SymmetrizedModel(
+                _EmptyModel(),
+                max_angular_momentum_target=0,
+                max_angular_momentum_character=2,
+                max_angular_momentum_grid=3,
+            )
+
+    @pytest.mark.parametrize(
```
Luthaf: I would rather have multiple tests one after the other instead of using a very large and hard to read parametrization

python/metatomic_torch/tests/symmetrized_model.py
```diff
@@ -0,0 +1,3101 @@
...
+        )
+
+    @pytest.mark.parametrize(
+        "source_name",
+        [
+            "energy/pbe",
+            "mtt::feature::node",
+            # "o3::variance_extra::" is not the reserved prefix: the full name
+            # is passed through as a source output
+            "o3::variance_extra::energy",
```
Luthaf: I think it would make more sense to reject anything starting with `o3::`, since we might want to add more outputs to the Symmetrized model in the future

python/metatomic_torch/tests/symmetrized_model.py
```diff
@@ -0,0 +1,3101 @@
...
+            # the reserved prefix is stripped exactly once, keeping "mtt::aux::"
```
Luthaf: not sure what this comment is trying to say

python/metatomic_torch/tests/symmetrized_model.py
```diff
@@ -0,0 +1,3101 @@
...
+        )
+
+    def test_rejects_invalid_requests_before_model_evaluation(self):
+        """Invalid public requests should fail without running the source model."""
+        base_model = _CountingLinearEnergyModel()
+        model = SymmetrizedModel(base_model, max_angular_momentum_target=0)
+        system = _forward_test_system([[1.0, 2.0, 3.0]])
+
+        assert model([], {}, None) == {}
+        message = "SymmetrizedModel requires at least one System"
```
Luthaf, line 1423: why? Can it not just return whatever the underlying returns in this case?

python/metatomic_torch/tests/symmetrized_model.py
```diff
@@ -0,0 +1,3101 @@
...
+        """Calling .float() on the module must fail loudly at the next forward."""
+        model = SymmetrizedModel(
+            _LinearEnergyModel(),
+            max_angular_momentum_target=0,
+            max_angular_momentum_grid=2,
+        ).float()
+
+        message = (
+            "SymmetrizedModel integration buffers must remain float64, got "
+            "torch.float32; do not call .float() or .half() on the module"
```
Luthaf: Why? If someone wants to use float32, it's on them. We can send a warning if it is really badly behaved in this case, but I can't see why it would?

python/metatomic_torch/tests/symmetrized_model.py
```diff
@@ -0,0 +1,3101 @@
...
+        """The complete forward path should execute after scripting and reloading."""
+        constructor_arguments = {
+            "max_angular_momentum_target": 0,
+            "max_angular_momentum_character": 1,
+            "max_angular_momentum_grid": 2,
+            "batch_size": 5,
+        }
+        eager = SymmetrizedModel(_LinearEnergyModel(), **constructor_arguments)
+        scripted = torch.jit.script(
+            SymmetrizedModel(_LinearEnergyModel(), **constructor_arguments)
```
Luthaf, line 1527: should this also test creating a SymmetrizedModel using a scripted + loaded inner model?

python/metatomic_torch/tests/symmetrized_model.py
```diff
@@ -0,0 +1,3101 @@
...
+                "" if source_output.unit == "" else f"({source_output.unit})^2"
+            )
+            assert capabilities.outputs["o3::variance::" + name].unit == squared_unit
+            character_name = "o3::character_projection::" + name
+            if max_angular_momentum_character is None:
+                assert character_name not in capabilities.outputs
+            else:
+                assert capabilities.outputs[character_name].unit == squared_unit
+
+    @pytest.mark.parametrize(
```
Luthaf: this giant parametrize is not very easy to read

python/metatomic_torch/tests/symmetrized_model.py
```diff
@@ -0,0 +1,3101 @@
...
+        )
+
+        message = (
+            "SymmetrizedModel supports CPU and CUDA execution, but the "
+            "wrapped model declares ['mps']"
+        )
+        with pytest.raises(ValueError, match=f"^{re.escape(message)}$"):
+            SymmetrizedModel.wrap(base, max_angular_momentum_target=0)
+
+    def test_preserves_requirements_and_runs_after_save_load(self, tmp_path):
```
Luthaf, line 1724: not sure what this is testing? It does not seems related to the SymmetrizedModel

python/metatomic_torch/tests/symmetrized_model.py
```diff
@@ -0,0 +1,3101 @@
...
+        assert torch.allclose(
+            reloaded["o3::variance::mtt::linear"].block().values,
+            expected_variance,
+            rtol=2.0e-5,
+            atol=1.0,
+        )
+
+    @pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA is not available")
+    def test_saved_wrapper_runs_on_cuda(self, tmp_path):
+        """Match CPU results after moving a saved float32 wrapper to CUDA."""
```
Luthaf, line 1846: isn't a test above checking that the model is not moved to float32??

python/metatomic_torch/tests/symmetrized_model.py
```diff
@@ -0,0 +1,3101 @@
...
+    samples = Labels(["system", "atom"], torch.tensor(sample_values))
+    block = TensorBlock(
+        values=torch.zeros((len(samples), 1), dtype=torch.float64),
+        samples=samples,
+        components=[],
+        properties=Labels.range("property", 1),
+    )
+
+    with pytest.raises(ValueError, match=f"^{re.escape(message)}$"):
+        group_samples_by_rotated_copy(block, n_rotated_copies=2)
```
Luthaf: seems to be testing internal functions, can this happen in practice?

python/metatomic_torch/tests/symmetrized_model.py
```diff
@@ -0,0 +1,3101 @@
...
+    assert torch.equal(reference.block().values, initial_reference_values)
+
+    # Simulate a later batch with the same layout but different response values.
+    tensor.block().values.add_(10.0)
+    later_values_by_copy = tensor.block().values.reshape(
+        n_rotated_copies, n_samples, *component_shape, n_properties
+    )
+    later_centered = later_values_by_copy - initial_reference_values.unsqueeze(0)
+
+    later_moments = _reduce_weighted_centered_batch(
```
Luthaf: private API again

Luthaf (commented):
A lot of tests are explicitly testing the private API. I would rather not (it makes changing the implementation a lot harder) unless there is a good reason for it.

We can also likely reduce the number of tests by re-using the O3 transformation code & tests.

python/metatomic_torch/metatomic/torch/o3/_symmetrized.py
ppegolo: Yes, but we might want to get rid of this? What about older checkpoints with old names?

python/metatomic_torch/metatomic/torch/o3/_symmetrized.py
ppegolo: It's for when deprecated names are used and normalized to the current names. It rejects spurious duplicates. Again, possibly overengineered

python/metatomic_torch/metatomic/torch/o3/_symmetrized.py
ppegolo: Sure. It's to avoid E[X²] − E[X]² being the difference of two (possibly) huge numbers. Subtracting a reference makes them always "small", and the difference is numerically stable

python/metatomic_torch/metatomic/torch/o3/_symmetrized.py
ppegolo: because we "have" to use float64 to guarantee numerical stability, and MPS doesn't support it. It prevents weirder errors down the line, but we can get rid of it, if you think it's too much

python/metatomic_torch/metatomic/torch/model.py
ppegolo: Yes, for torchscript

python/metatomic_torch/metatomic/torch/o3/_decompose.py
ppegolo: More because this is for symmetric matrices and there it's zero by construction, but it might actually make sense to compute it as diagnostic, since PET's `non_conservative_stress` is not forced to be symmetric within the model, I think

python/metatomic_torch/tests/symmetrized_model.py
ppegolo: Not right now, as the symmetrized model infers some metadata after evaluation on systems. If there's a clean way to avoid that we can do it though

python/metatomic_torch/metatomic/torch/o3/_symmetrized.py
Luthaf: I think older checkpoints already get renamed in metatrain checkpoint upgrade functions. If not they should!

python/metatomic_torch/tests/symmetrized_model.py
ppegolo: Right, allowed now. In tests we made the final numbers changed quite significantly though, but that was possibly before the robust variance estimation

python/metatomic_torch/metatomic/torch/o3/_decompose.py
Luthaf: would it make sense to compute it and warn if it is bigger than some threshold? Or would this be just noise?

python/metatomic_torch/metatomic/torch/model.py
Luthaf: maybe copy.deepcopy is clearer then

python/metatomic_torch/metatomic/torch/o3/_symmetrized.py
Luthaf: ah, makes sense. Then just some update in the docstring would be nice!

python/metatomic_torch/tests/symmetrized_model.py
ppegolo: It's the integration buffers that must stay (or had do, before allowing float32 as well) float64. But it's true that here it was unclear

python/metatomic_torch/tests/symmetrized_model.py
ppegolo: It's testing the wrapper but might be out of scope yes

python/metatomic_torch/metatomic/torch/o3/_decompose.py
ppegolo: I think it might be "big" in general, so maybe it's good to compute it all the time

Luthaf:
merged in #300

PR #296, opened 2026-07-30
fix float max_neighbors passed to nvalchemi

joehart2001:
# Summary

Fix a type error in the CUDA/nvalchemi neighbour-list backend for large cutoffs.

#259 added a `max_neighbors` estimate:

```python
len(system) * max(128, cutoff**3)
```
`cutoff` is a float, so this expression is also a float once `cutoff**3 > 128` (i.e. above roughly 5 Å). However, nvalchemi requires `max_neighbors` to be an integer.

I noticed this whilst running benchmarks on ml-peg with pet-cam-xl. this can be reproduced by running e.g. the lattice constant benchmark on gpu.

# Reviewer checklist

### DIFF
--- a/python/metatomic_ase/src/metatomic_ase/_neighbors.py
+++ b/python/metatomic_ase/src/metatomic_ase/_neighbors.py
@@ -1,3 +1,4 @@
+import math
 import warnings
 from typing import List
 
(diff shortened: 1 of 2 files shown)


frostedoyster (approved):
Thanks!

PR #306, opened 2026-08-31
Implement `mta_execute_model`

Luthaf:
This should be the last building block for the new C API!

# Reviewer checklist

### DIFF
--- a/metatomic-core/src/kernels/cuda.rs
+++ b/metatomic-core/src/kernels/cuda.rs
@@ -127,14 +127,8 @@ pub(crate) fn is_equal_i32(tensor: DLPackTensorRef<'_>, reference: &ReferenceVal
     // Wrap the existing GPU-allocated tensor pointer
     let tensor_ptr = unsafe { DevicePtrArg { ptr: dlpack_to_device_ptr(&tensor) } };
 
-    // Upload reference values to GPU (cached after first call)
-    let (ref_dev, reference_idx) = reference.cuda.get_or_init(|| {
-        let slice = stream
-            .clone_htod(reference.cpu.as_slice().expect("reference should be contiguous"))
-            .expect("clone_htod reference failed");
-        let idx = StridedNDIndex::from_ndarray(&reference.cpu.view());
-        (slice, idx)
-    });
+    // Upload reference values to GPU (cached after first call, per device)
+    let (ref_dev, reference_idx) = reference.cuda_data(device_id, &stream)?;
 
     // Allocate result flag (initialized to 0 = no mismatch)
     let mut result = stream.alloc_zeros::<i32>(1)
--- a/metatomic-core/src/kernels/cuda.rs
+++ b/metatomic-core/src/kernels/cuda.rs
@@ -232,3 +232,157 @@ pub(crate) fn validate_cell_pbc(
     }
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use dlpk::{DLDevice, GetDLPackDataType};
+    use ndarray::ArrayD;
+
+    /// Check whether a CUDA device is available on this machine. The tests
+    /// below are skipped when there is none.
+    fn cuda_available() -> bool {
+        // this requires the `dynamic-loading` feature of cudarc, without which
+        // the tests would fail to link on machines without CUDA anyway
+        if !unsafe {cudarc::driver::sys::is_culib_present() } {
+            return false;
+        }
+        return CudaContext::device_count().unwrap_or(0) > 0;
+    }
+
+    macro_rules! skip_without_cuda {
+        () => {
+            if !cuda_available() {
+                eprintln!("no CUDA device available, skipping this test");
+                return;
+            }
+        };
+    }
+
+    /// A DLPack tensor with CUDA-resident data, used to test the kernels above.
+    struct CudaTensor {
+        ptr: cudarc::driver::sys::CUdeviceptr,
+        shape: Vec<i64>,
+        strides: Vec<i64>,
+        dtype: dlpk::sys::DLDataType,
+    }
+
+    impl CudaTensor {
+        /// Create a new CUDA tensor with the given `shape` and `strides`,
+        /// containing a copy of `data`.
+        ///
+        /// `data` is the full memory span of the tensor, including any gap
+        /// between the elements actually part of the tensor.
+        fn new<T: GetDLPackDataType + Copy>(data: &[T], shape: &[i64], strides: &[i64]) -> Self {
+            let stream = get_or_init(0).expect("failed to initialize CUDA device 0");
+            stream.context().bind_to_thread().expect("bind_to_thread failed");
+
+            assert!(!data.is_empty());
+            let ptr = unsafe {
+                cudarc::driver::result::malloc_sync(std::mem::size_of_val(data))
+            }.expect("malloc_sync failed");
+
+            if !data.is_empty() {
+                unsafe {
+                    cudarc::driver::result::memcpy_htod_sync(ptr, data)
+                }.expect("memcpy_htod_sync failed");
+            }
+
+            CudaTensor {
+                ptr,
+                shape: shape.to_vec(),
+                strides: strides.to_vec(),
+                dtype: T::get_dlpack_data_type(),
+            }
+        }
+
+        #[allow(clippy::cast_possible_truncation, clippy::cast_possible_wrap)]
+        fn as_ref(&self) -> DLPackTensorRef<'_> {
+            unsafe {
+                DLPackTensorRef::from_raw(dlpk::sys::DLTensor {
+                    data: self.ptr as *mut std::ffi::c_void,
+                    device: DLDevice {
+                        device_type: dlpk::sys::DLDeviceType::kDLCUDA,
+                        device_id: 0,
+                    },
+                    ndim: self.shape.len() as i32,
+                    dtype: self.dtype,
+                    shape: self.shape.as_ptr().cast_mut(),
+                    strides: self.strides.as_ptr().cast_mut(),
+                    byte_offset: 0,
+                })
+            }
+        }
+    }
+
+    impl Drop for CudaTensor {
+        fn drop(&mut self) {
+            unsafe {
+                let _ = cudarc::driver::result::free_sync(self.ptr);
+            }
+        }
+    }
+
+    #[test]
+    fn is_equal_i32_kernel() {
+        skip_without_cuda!();
+
+        let reference = ReferenceValue::new(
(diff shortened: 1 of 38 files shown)


metatomic-core/src/model/execute.rs
```diff
@@ -0,0 +1,234 @@
...
+            systems,
+            selected_atoms_labels.as_ref(),
+        )?;
+    }
+
+    // always check atomic types (even when check_consistency is false)
+    {
+        let mut cache = ATOMIC_TYPES_CACHE.lock().expect("ATOMIC_TYPES_CACHE lock poisoned");
+        let atomic_types = &capabilities.atomic_types;
+        let valid_types = cache.get_or_insert_ref(atomic_types.as_slice(), || {
```
RMeli: Is `atomic_types` guaranteed to be sorted? (From the comment, you want the key to be sorted.)

metatomic-core/src/model/inputs.rs
```diff
@@ -0,0 +1,482 @@
...
+            }
+        }
+    }
+
+    Ok(())
+}
+
+/// Validate that the inputs to a model are consistent with its capabilities.
+///
+/// This checks that:
```
RMeli, line 67: Maybe worth checking also that the device is in the model's supported devices?

metatomic-core/src/utils.rs
```diff
@@ -0,0 +1,151 @@
...
+                &samples,
+                &components,
+                &properties,
+            )?;
+
+            // copy and scale all gradients
+            for (parameter, gradient) in block.gradients() {
+                let grad_device = gradient.values().device()?;
+                let grad_copy = gradient.values().copy(grad_device)?;
+                let mut dlpack = grad_copy.as_dlpack(device, None, dlpack_version)?;
```
RMeli, line 80:
```suggestion
                let mut dlpack = grad_copy.as_dlpack(grad_device, None, dlpack_version)?;
```

metatomic-core/src/kernels/cuda.rs
```diff
@@ -238,3 +293,255 @@ pub(crate) fn validate_cell_pbc(
...
+pub(super) fn clone_tensor(tensor: &DLPackTensorRef<'_>) -> Result<DLPackTensor, Error> {
+    check_valid_device("clone_tensor", tensor.device());
+
+    let device_id = tensor.device().device_id as usize;
+    let stream = get_or_init(device_id)?;
+
+    let src_ptr = unsafe { dlpack_to_device_ptr(*tensor) };
+    let n_elements: i64 = tensor.shape().iter().product();
+    let elem_size = tensor.dtype().bits as usize / 8;
+    let num_bytes = n_elements as usize * elem_size;
```
RMeli: Worth checking that the tensor is contiguous?

metatomic-core/src/model/execute.rs
```diff
@@ -0,0 +1,234 @@
...
+pub fn execute_model(
+    model: &Model,
+    systems: &[Arc<System>],
+    selected_atoms: *const mts_labels_t,
+    requested_outputs_json: &str,
+    check_consistency: bool,
+    outputs: *mut *mut mts_tensormap_t,
+    outputs_count: usize,
+) -> Result<(), Error> {
+    let capabilities = model.capabilities()?;
```
RMeli, line 61: Can these be cached? Doing string allocation, JSON conversion, etc at every call might be wasteful.

metatomic-core/src/model/execute.rs
Luthaf: no, I forgot to sort them in there. I've made this part of the type contract so we don't get it wrong next time.

metatomic-core/src/model/execute.rs
Luthaf: I think we can try to get them at every call for now, and I'll leave a TODO note to measure the overhead of this

RMeli: Sounds good.

metatomic-core/src/kernels/cuda.rs
Luthaf: I changed the kernel to instead clone to a new contiguous array, and use the existing ND indexing facilities

Luthaf:
Ok, so I ended up adding tests for the cuda and metal kernels, and this found quite a few bugs. Everything should now be fixed and ready for a second review. I'll add a TODO item to run CI on real GPU hardware for a follow-up

metatomic-core/src/kernels/cuda.rs
```diff
@@ -238,3 +312,624 @@ pub(crate) fn validate_cell_pbc(
...
+
+    let device_id = tensor.device().device_id as usize;
+    let stream = get_or_init(device_id)?;
+
+    let cache = CUDA_CACHE.lock().expect("failed to lock CUDA_CACHE");
+    let entry = &cache[&device_id];
+
+    let element_size = super::element_size(tensor.dtype())?;
+    let n_elements: i64 = tensor.shape().iter().product();
+    let num_bytes = n_elements as usize * element_size;
```
RMeli: In the metal kernel you have
```
    let element_size = super::element_size(tensor.dtype())?;
    let n_elements: usize = tensor.shape().iter().map(|&s| s as usize).product();
    let num_bytes = n_elements * element_size / 8;
```

1. Is `element_size` in bits or bytes?
2. The version for `n_elements` here is a bit cleaner, use the same (i.e. no `map`) in the metal kernel?

metatomic-core/src/kernels/metal.rs
```diff
@@ -309,3 +406,615 @@ pub(crate) fn validate_cell_pbc(
...
+#[allow(clippy::cast_sign_loss, clippy::cast_possible_truncation, clippy::cast_possible_wrap)]
+pub(super) fn clone_tensor(tensor: &DLPackTensorRef<'_>) -> Result<DLPackTensor, Error> {
+    check_valid_device("clone_tensor", tensor.device());
+
+    let device_id = tensor.device().device_id as usize;
+    let mut lock = METAL_CACHE.lock().expect("failed to lock METAL_CACHE");
+    let cache = get_or_init(&mut lock, device_id)?;
+
+    let element_size = super::element_size(tensor.dtype())?;
+    let n_elements: usize = tensor.shape().iter().map(|&s| s as usize).product();
```
RMeli, line 625: See comment above.

metatomic-core/src/kernels/cuda.rs
RMeli: The size is actually in bits, so this seems wrong (allocating 8x). Might be worth renaming to `element_size_bits`/`element_size_bytes` to avoid this kind of error?

metatomic-core/src/model/execute.rs
```diff
@@ -0,0 +1,254 @@
...
+        let mut cache = ATOMIC_TYPES_CACHE.lock().expect("ATOMIC_TYPES_CACHE lock poisoned");
+        let atomic_types = AtomicTypes::new(capabilities.atomic_types.clone());
+        let valid_types = cache.get_or_insert(atomic_types, || {
+            ReferenceValue::new(
+                ndarray::ArrayD::from_shape_vec(
+                    ndarray::IxDyn(&[capabilities.atomic_types.len()]),
+                    capabilities.atomic_types.clone(),
+                ).expect("atomic_types should be contiguous")
+            )
+        });
```
RMeli, line 124: Release the lock here, before looping over systems?

metatomic-core/src/kernels/metal.rs
```diff
@@ -309,3 +406,615 @@ pub(crate) fn validate_cell_pbc(
...
+    // Upload valid types to Metal (cached after first call, per device)
+    let (valid_types_buffer, _) = valid_types.metal_data(device_id, &cache.device)?;
+    let n_valid_types = valid_types.cpu.len() as u64;
+
+    let types_buf = unsafe {
+        cache.device.newBufferWithBytes_length_options(
+            NonNull::new(dlpack_data_ptr(types).cast_mut()).expect("types pointer must not be null"),
+            dlpack_num_bytes(types),
+            MTLResourceOptions::empty(),
+        ).expect("failed to create types buffer")
```
RMeli: I might be missing something, but why is this used instead of the `MetalBufferRef::from_dlpack` used above?

metatomic-core/src/model/execute.rs
```diff
@@ -0,0 +1,254 @@
...
+            // free any outputs that were filled so far
+            for j in 0..outputs_count {
+                unsafe {
+                    let ptr = std::mem::replace(&mut *outputs.add(j), std::ptr::null_mut());
+
+                    if !ptr.is_null() {
+                        std::mem::drop(TensorMap::from_raw(ptr));
+                    }
+                }
+            }
```
RMeli: Same code as above. It is not too long, but what about using an helper function instead?

RMeli:
Great you added Metal and CUDA tests and squashed most bugs!

> TODO item to run CI on real GPU hardware for a follow-up

Yay!

metatomic-core/src/kernels/cuda.rs
Luthaf: good point, this is a partial refactor issue =\

metatomic-core/src/model/execute.rs
Luthaf: we can't because if we release the lock some other thread could write to it and invalidate the reference we hold to `valid_types` (it would not even compile)

metatomic-core/src/kernels/metal.rs
Luthaf: no real reason, just a copy-paste. Will update

metatomic-core/src/model/execute.rs
RMeli: Right, I missed it was a reference. This serializes all the kernel launches afterwards though.

Luthaf: Which kernel launches? The loop over systems should be able to run in parallel (they don't currently, it would require a refactor to use futures here), and then we release the lock when we exit the scope line 128

RMeli: The `check_atomic_types`; they are serial in the loop, but also serialized by the lock with other calls to `execute_model`.

Luthaf: ok, right. 

> but also serialized by the lock with other calls to execute_model.

this we can't remove because of the reference thing from above. It should also not matter that much unless a piece of code is trying to run many simulations on different set of systems (and thus with different models) at the same time. If we really need to get there we could maybe trade off by making the `valid_types` reference counted instead, but I'd leave this for later.

> they are serial in the loop

This is the bit we could fix by making all kernels async functions, spawning them here and joining later. Also something I'd leave for later though, even if it could be pretty fun to do ^_^

RMeli (approved):

PR #314, opened 2026-09-03
C++ API for custom models

RMeli:
`BaseModel` can be used as base class to define custom models in C++. `ExternalModel` can be used to wrap existing models.

[📚 Download documentation for this pull-request](https://nightly.link/metatensor/metatomic/actions/artifacts/10048545394.zip)

[⚙️ Download Python wheels for this pull-request (you can install these with pip)](https://nightly.link/metatensor/metatomic/actions/artifacts/9895934964.zip)


metatomic-core/include/metatomic/model.hpp
```diff
@@ -19,4 +26,353 @@ namespace metatomic {
...
+        virtual ModelCapabilities capabilities() const = 0;
+
+        /// Get metadata describing this model.
+        virtual ModelMetadata metadata() const = 0;
+
+        /// List the outputs this model is able to compute.
+        ///
+        /// The default implementation returns the outputs declared in
+        /// `capabilities()`.
+        virtual std::vector<Quantity> supported_outputs() const {
```
Luthaf, line 49: I am now wondering why we have this information duplicated in two separate places … It might make sense to keep only one of them. Since `mta_execute_model` calls both, and most engine would also need both at the same time, we might be better off removing supported_outputs everywhere. What do you think?

metatomic-core/include/metatomic/model.hpp
```diff
@@ -19,4 +26,353 @@ namespace metatomic {
...
+            return capabilities().outputs();
+        }
+
+        /// List the pair lists (neighbor lists) this model needs as input.
+        virtual std::vector<PairListOptions> requested_pair_lists() const = 0;
+
+        /// List the additional per-system inputs this model needs.
+        virtual std::vector<Quantity> requested_inputs() const = 0;
+
+        /// Run the model and compute the requested outputs.
```
Luthaf, line 59: Maybe a comment here that this should not be called directly but rather through `mta_execute_model`. Maybe naming it something like `execute_inner` or similar to make API reader have a double take?

metatomic-core/include/metatomic/model.hpp
```diff
@@ -19,4 +26,353 @@ namespace metatomic {
...
+        ) = 0;
+
+        /// Convert a C++ model to a `mta_model_t`.
+        ///
+        /// The returned `mta_model_t` takes ownership of the model and will
+        /// delete it when the `unload` callback is called.
+        ///
+        /// @param model model to convert
+        /// @return a `mta_model_t` model
+        static mta_model_t to_mta_model(std::unique_ptr<BaseModel> model) {
```
Luthaf: can we specialize this for `ExternalModel`?

metatomic-core/tests/cxx/model.cpp
```diff
@@ -0,0 +1,323 @@
...
+                "test-cpp-plugin",
+                load_cpp_model,
+            };
+            auto status = mta_register_plugin(PLUGIN);
+            if (status != MTA_SUCCESS) {
+                throw metatomic::Error("failed to register test-cpp-plugin");
+            }
+        }
+    } CPP_PLUGIN_REGISTRAR;
+}
```
Luthaf: I'd leave this and related tests to a follow-up PR handling plugins in C++ (and then we can build a C++ plugin for the tests, like we do for the C plugin)

metatomic-core/include/metatomic/model.hpp
RMeli: Isn't this just a convenience function? Maybe it should be non-virtual and always return `capabilities().outputs()`. 

I agree it might not be that useful to have both. Should I remove it then, and we work directly with capabilities?

Luthaf: I'd say we remove it from the C API in a separate PR and then update this one. We can add it back as a convenience down the line if it is useful.

Luthaf (approved):
Looks good, we can do the `supported_outputs` removal first in a separate PR, or merge this one and then remove `supported_outputs`, as you prefer!

RMeli:
Let's merge first, then I'll open the removal PR.

PR #316, opened 2026-09-04
Update metatomic-torchsim to vesin 0.6

### LINKED ISSUE #315: metatomic-torchsim's vesin pin (<0.6) conflicts with metatomic-ase's (>=0.6), making metatomic[torch,torchsim] unresolvable
`metatomic-torch` depends on `metatomic-ase`, which requires `vesin>=0.6.0,<0.7` (since #283, "Update to vesin v0.6"). `metatomic-torchsim` still pins `vesin>=0.5.6,<0.6` (python/metatomic_torchsim/setup.py:110) *and* depends on `metatomic-torch` itself, so the two requirements are unsatisfiable together -- confirmed this is not an extras-combination artifact by installing `metatomic[torchsim]` completely alone in an empty venv and getting the identical resolver error.

Repro:
```
uv pip install -e "./metatomic[torchsim]"
```
```
  x No solution found when resolving dependencies:
  |-> Because only metatomic-ase==0.1.4 is available and metatomic-ase==0.1.4
      depends on vesin>=0.6.0,<0.7, we can conclude that all versions of
      metatomic-ase depend on vesin>=0.6.0,<0.7.
      And because metatomic-torch==0.1.17 depends on metatomic-ase and only
      metatomic-torch==0.1.17 is available, we can conclude that all versions
      of metatomic-torch depend on vesin>=0.6.0,<0.7.
      And because metatomic-torchsim==0.1.6 depends on vesin>=0.5.6,<0.6 and
      metatomic-torch, we can conclude that metatomic-torchsim==0.1.6 cannot
      be used.
```

Looks like the `vesin` bump in #283 just missed updating the pin in `metatomic_torchsim/setup.py`. Bumping it to match (or dropping the upper bound) should fix it.

EricBoittier:
This was missed in #283, so `metatomic[torchsim]` could not resolve: `metatomic-torchsim` required `vesin<0.6` while `metatomic-ase` requires `vesin>=0.6`.

Fixes #315.

 - [ ] ~Tests updated (for new features and bugfixes)?~

# Reviewer checklist

### DIFF
--- a/python/metatomic_torchsim/CHANGELOG.md
+++ b/python/metatomic_torchsim/CHANGELOG.md
@@ -7,6 +7,10 @@ follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased](https://github.com/metatensor/metatomic/)
 
+### Changed
+
+- Update the `vesin` dependency to `>=0.6.0,<0.7`, matching `metatomic-ase`.
+
 <!-- Possible sections
 ### Added
 
--- a/python/metatomic_torchsim/setup.py
+++ b/python/metatomic_torchsim/setup.py
@@ -107,7 +107,7 @@ if __name__ == "__main__":
 
     install_requires = [
         "torch-sim-atomistic >=0.5",
-        "vesin >=0.5.6,<0.6",
+        "vesin >=0.6.0,<0.7",
     ]
 
     # when packaging a sdist for release, we should never use local dependencies


Luthaf:
Please keep the existing PR template when opening PR. Also, if you have a test plan in there, please do run it and check the corresponding check boxes.

Luthaf (approved):

PR #318, opened 2026-09-04
Add a WeightedSum model wrapper

zyxwwxyz:
Adds `metatomic.torch.weighted_sum.WeightedSum`, a wrapper that combines several
existing outputs of a model into a new output computing their fixed linear
combination (for example, combining several `"energy/<head>"` outputs, such as
predictions from different DFT functionals, into a single `"energy"` output).

All the heads entering a given weighted sum are requested from the wrapped
model in a single call and combined by directly operating on the raw
`TensorMap` values (`mts.add`/`mts.multiply`), without detaching or otherwise
interrupting the autograd graph. This means a single `backward()` call through
the weighted-sum output produces forces and stresses that are exactly the
weighted sum of the forces and stresses of the individual heads, instead of
requiring one evaluation (and one set of gradients) per head.

The wrapper follows the same `wrap()`-based pattern as `HeatFlux` and `DFTD3`.
An optional `normalize_coefficients` flag on `wrap()` rescales the given
coefficients to sum to one (this also works with negative coefficients, as
long as they don't sum to exactly zero).

This is motivated by checkpoints exposing several "proper" heads (e.g. one per
training-data functional) that should be combined into physically meaningful
targets - for example an ensemble/consensus energy, without paying for a
separate forward+backward evaluation per combination.

# Reviewer checklist

### DIFF
--- a/metatomic-torch/CHANGELOG.md
+++ b/metatomic-torch/CHANGELOG.md
@@ -6,6 +6,19 @@ a changelog](https://keepachangelog.com/en/1.1.0/) format. This project follows
 
 ## [Unreleased](https://github.com/metatensor/metatomic/)
 
+### Added
+
+- Added `metatomic.torch.weighted_sum.WeightedSum`, a wrapper computing a fixed
+  linear combination of several existing outputs of a model (for example
+  combining several `"energy/<head>"` outputs into a single `"energy"`
+  output) in a single evaluation of the wrapped model, preserving the
+  autograd graph so that forces and stresses of the combined output are
+  exactly the weighted sum of the forces and stresses of the individual
+  heads. `WeightedSum.wrap` accepts a `normalize_coefficients` flag to rescale
+  the given coefficients so they sum to one (including when some of them are
+  negative); a warning is emitted if the sum of coefficients is negative,
+  since normalizing then flips the sign of every coefficient.
+
 <!-- Possible sections:
 ### Added
 
--- a/python/metatomic_torch/metatomic/torch/weighted_sum.py
+++ b/python/metatomic_torch/metatomic/torch/weighted_sum.py
@@ -1,0 +1,279 @@
+"""
+:py:class:`WeightedSum`, which wraps a model exposing several "proper" head outputs
+for the same physical quantity (for example several energies computed with
+different functionals, such as ``"energy/pbe"``, ``"energy/r2scan"`` and
+``"energy/lda"``) and adds a new output computing a fixed linear combination of
+these heads.
+"""
+
+import warnings
+from typing import Dict, List, Optional
+
+import metatensor.torch as mts
+import torch
+from metatensor.torch import Labels, TensorMap
+
+from . import (
+    ModelCapabilities,
+    ModelOutput,
+    NeighborListOptions,
+    System,
+)
+from .model import (
+    AtomisticModel,
+    ModelInterface,
+)
+
+
+class WeightedSum(torch.nn.Module):
+    """
+    Wrap a model with a new output computing a fixed weighted sum of several of its
+    existing outputs.
+
+    The proper heads combined in the sum (for example ``"energy/pbe"``,
+    ``"energy/r2scan"`` and ``"energy/lda"``) are all requested from the wrapped
+    model in a single call, and the weighted-sum output is built by directly
+    combining the resulting :py:class:`TensorMap` values with :py:func:`mts.add` and
+    :py:func:`mts.multiply`. This does not evaluate the model more than once and does
+    not detach or otherwise interrupt the autograd graph, so the weighted-sum output
+    stays connected to the same computational graph as each individual head. If the
+    input systems require gradients, a single ``backward()`` call through the
+    weighted-sum output therefore produces forces and stresses that are exactly the
+    weighted sum of the forces and stresses of the individual heads.
+
+    :param model: underlying :py:class:`ModelInterface`. The :py:meth:`wrap` method
+        obtains this module from :py:attr:`AtomisticModel.module`.
+    :param output_name: name of the new output computing the weighted sum, e.g.
+        ``"energy"``.
+    :param weights: mapping from the name of an existing head output of ``model``
+        (e.g. ``"energy/pbe"``) to its fixed coefficient in the weighted sum.
+    """
+
+    _output_name: str
+    _weights: Dict[str, float]
+    _requested_inputs: Dict[str, ModelOutput]
+    _requested_neighbor_lists: List[NeighborListOptions]
+
+    def __init__(
+        self,
+        model: ModelInterface,
+        output_name: str,
+        weights: Dict[str, float],
+    ):
+        super().__init__()
+
+        if len(weights) == 0:
+            raise ValueError(
+                "`weights` must contain at least one head output name to combine"
+            )
+
+        self._model = model
+        self._output_name = output_name
+        self._weights = weights
+        self._requested_inputs = {}
+        self._requested_neighbor_lists = []
+
+    @staticmethod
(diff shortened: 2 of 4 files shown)


metatomic-torch/CHANGELOG.md
```diff
@@ -6,6 +6,19 @@ a changelog](https://keepachangelog.com/en/1.1.0/) format. This project follows
...
+- Added `metatomic.torch.weighted_sum.WeightedSum`, a wrapper computing a fixed
+  linear combination of several existing outputs of a model (for example
+  combining several `"energy/<head>"` outputs into a single `"energy"`
+  output) in a single evaluation of the wrapped model, preserving the
+  autograd graph so that forces and stresses of the combined output are
+  exactly the weighted sum of the forces and stresses of the individual
+  heads. `WeightedSum.wrap` accepts a `normalize_coefficients` flag to rescale
+  the given coefficients so they sum to one (including when some of them are
+  negative); a warning is emitted if the sum of coefficients is negative,
+  since normalizing then flips the sign of every coefficient.
```
Luthaf:
```suggestion
- Added `metatomic.torch.weighted_sum.WeightedSum`, a wrapper computing a fixed
  linear combination of several existing outputs of a model, for example
  combining several `"energy/<head>"` outputs into a single `"energy"`
  output.
```

Too many details for a CHANGELOG =)

python/metatomic_torch/metatomic/torch/weighted_sum.py
```diff
@@ -0,0 +1,279 @@
+"""
+:py:class:`WeightedSum`, which wraps a model exposing several "proper" head outputs
+for the same physical quantity (for example several energies computed with
+different functionals, such as ``"energy/pbe"``, ``"energy/r2scan"`` and
+``"energy/lda"``) and adds a new output computing a fixed linear combination of
+these heads.
+"""
```
Luthaf, line 7: does the code handles the case where the wrapped model has two outputs, one without variant? (`energy` and `energy/pbe`)?

In general, are the original variants still accessible in some way?

python/metatomic_torch/metatomic/torch/weighted_sum.py
```diff
@@ -0,0 +1,279 @@
...
+    :param output_name: name of the new output computing the weighted sum, e.g.
+        ``"energy"``.
+    :param weights: mapping from the name of an existing head output of ``model``
+        (e.g. ``"energy/pbe"``) to its fixed coefficient in the weighted sum.
+    """
+
+    _output_name: str
+    _weights: Dict[str, float]
+    _requested_inputs: Dict[str, ModelOutput]
+    _requested_neighbor_lists: List[NeighborListOptions]
```
Luthaf: why these? metatomic should already recursively search sub-module for the corresponding functions and respect them.

python/metatomic_torch/metatomic/torch/weighted_sum.py
```diff
@@ -0,0 +1,279 @@
...
+            raise TypeError("model must be an AtomisticModel")
+
+        if len(weights) == 0:
+            raise ValueError(
+                "`weights` must contain at least one head output name to combine"
+            )
+
+        if normalize_coefficients:
+            coefficients_sum = sum(weights.values())
+            if coefficients_sum == 0:
```
Luthaf:
```suggestion
            if abs(coefficients_sum) < 1e-6:
```

python/metatomic_torch/metatomic/torch/weighted_sum.py
```diff
@@ -0,0 +1,279 @@
...
+        """Evaluate the wrapped model and the requested weighted sum."""
+        if self._output_name not in outputs:
+            return self._model(systems, outputs, selected_atoms)
+
+        requested = outputs[self._output_name]
+        if len(requested.explicit_gradients) != 0:
+            raise ValueError(
+                "WeightedSum does not support explicit gradients for the "
+                f"'{self._output_name}' output"
+            )
```
Luthaf: This should already be empty if declared empty in the capabilities

python/metatomic_torch/tests/weighted_sum.py
```diff
@@ -0,0 +1,822 @@
...
+    ModelMetadata,
+    ModelOutput,
+    NeighborListOptions,
+    System,
+    load_atomistic_model,
+)
+from metatomic.torch.weighted_sum import WeightedSum
+
+
+ATOMIC_NUMBER = 6
```
Luthaf: ?? Why is this a global constant?

python/metatomic_torch/tests/weighted_sum.py
```diff
@@ -0,0 +1,822 @@
...
+            length_unit="Angstrom",
+            supported_devices=["cpu", "cuda"],
+            dtype="float64",
+        ),
+    )
+
+
+@pytest.fixture
+def weights():
+    return {"energy/pbe": 0.6, "energy/r2scan": 0.3, "energy/lda": 0.1}
```
Luthaf: I'm not sure this is worth a fixture vs writing inline in the tests that need it

python/metatomic_torch/metatomic/torch/weighted_sum.py
zyxwwxyz: Yes to both. 
1) Original variants stay accessible. `wrap` copies every output declared by the wrapped model into the new capabilities, and `forward` passes through anything the caller requests other than the weighted sum itself. So `energy/pbe` & co. remain requestable, on their own or in the same call as the sum.
2) `energy` and `energy/pbe` coexisting is fine as long as the new output gets its own name. Wrapping such a model with `output_name="energy/mix"` gives capabilities `['energy', 'energy/mix', 'energy/pbe', 'energy/r2scan']`, all requestable in a single call, with the plain `energy` bit-identical to what the unwrapped model returns.

If you pass `output_name="energy"` for a model that already declares `energy`, you get `ValueError: this model already has an output named 'energy', which conflicts with the weighted-sum output`.

I've added `test_weighted_sum_with_plain_and_variant_outputs` and `test_weighted_sum_rejects_plain_output_name_conflict` covering both cases, and spelled this out in the `wrap` docstring. (see [7017281)](https://github.com/metatensor/metatomic/pull/318/changes/7017281150167503633a37cea1a3779406ebc011)

python/metatomic_torch/metatomic/torch/weighted_sum.py
zyxwwxyz: You are right, this was redundant. I removed it in 5a6e687.

python/metatomic_torch/metatomic/torch/weighted_sum.py
zyxwwxyz: You're right, a normal user will never incur into this `if` condition. I will cancel these lines and rewrite `test_weighted_sum_rejects_explicit_gradients`. A similar `if` lies in https://github.com/metatensor/metatomic/blob/3fea212674bb7c337283efdcb382cf976ead8124/python/metatomic_torch/metatomic/torch/o3/_symmetrized.py#L800-L805, don't know if we want to remove it as well

Luthaf: yes please

python/metatomic_torch/tests/weighted_sum.py
zyxwwxyz: Yes this is not nice

python/metatomic_torch/tests/weighted_sum.py
zyxwwxyz: Makes sense, I moved the definition of `weights` inside each test that needs it.

PR #266, opened 2026-06-23
C++ API to format model metadata

RMeli:
[📚 Download documentation for this pull-request](https://nightly.link/metatensor/metatomic/actions/artifacts/7816087042.zip)

[⚙️ Download Python wheels for this pull-request (you can install these with pip)](https://nightly.link/metatensor/metatomic/actions/artifacts/7816255910.zip)

PR #319, opened 2026-09-08
Use builders for metadata objects in C++

RMeli:
As discussed offline, I did not like our fist implementation relying on `std::optional` (see #266). This PR uses a builder/factory pattern instead, creating immutable metadata classes (immutability is a constraint that can be relaxed, if needed). I think this solution is much cleaner, and achieves the same goal.

[📚 Download documentation for this pull-request](https://nightly.link/metatensor/metatomic/actions/artifacts/10105039931.zip)


metatomic-core/tests/cxx/metadata.cpp
```diff
@@ -29,12 +35,14 @@ TEST_CASE("JSON serialization C++ API") {
...
-            p1.full_list(true);
-            p1.strict(false);
-            p1.requestors({"model1", "model2"});
+        SECTION("Builder with requestors set as a list") {
+            std::vector<std::string> requestors = {"model1", "model2"};
+            auto p1 = metatomic::PairListOptions::builder()
+                .cutoff(cutoff)
+                .full_list(true)
+                .strict(false)
+                .requestors(requestors)
```
Luthaf:
```suggestion
            auto p1 = metatomic::PairListOptions::builder()
                .cutoff(cutoff)
                .full_list(true)
                .strict(false)
                .requestors({"model1", "model2"})
```

Would this compile? it would be nice =)

metatomic-core/include/metatomic/metadata.hpp
```diff
@@ -41,89 +41,66 @@ namespace metatomic {
...
         std::vector<std::string> requestors_;
 
-    public:
-        /// Set the cutoff radius for this pair list.
+        /// Validate that `value` is a finite positive number
         ///
-        /// @throw metatomic::Error if the value is not a finite positive number.
-        void cutoff(double value) {
+        /// @throw metatomic::Error if `value` is not a finite positive number
+        static void validate_cutoff(double value) {
```
Luthaf: should the validation belong to the builder instead?

metatomic-core/include/metatomic/metadata.hpp
```diff
@@ -41,89 +41,66 @@ namespace metatomic {
...
+        /// Add `requestor` to `requestors_`, ignoring empty strings and
+        /// duplicates, keeping first-seen order.
         ///
-        /// @throw metatomic::Error if the value has not been set.
-        double cutoff() const {
-            if (!cutoff_.has_value()) {
-                throw metatomic::Error("cutoff is not set in PairListOptions");
+        /// @param requestors the list of requestors to add to
+        /// @param requestor the requestor to add
+        static void add_requestor_to(std::vector<std::string>& requestors, const std::string& requestor) {
```
Luthaf: same, should this be part of the builder instead?

metatomic-core/tests/cxx/metadata.cpp
RMeli: Yep! Changed.

metatomic-core/include/metatomic/metadata.hpp
RMeli: Yes, good catch!

metatomic-core/include/metatomic/metadata.hpp
RMeli: Yes, good catch!

Luthaf (approved):

PR #323, opened 2026-09-09
C++ wrapper for mta_execute_model

RMeli:
Implement `metatomic::execute_model` based on `mta_execute_model`, and uncomment the related tests.

[📚 Download documentation for this pull-request](https://nightly.link/metatensor/metatomic/actions/artifacts/10391132110.zip)

### DIFF
--- a/metatomic-core/include/metatomic/model.hpp
+++ b/metatomic-core/include/metatomic/model.hpp
@@ -356,4 +356,59 @@ namespace metatomic {
 
         return m;
     }
+
+    /// Execute a model to compute the requested outputs for a set of systems.
+    ///
+    /// @param model the model to execute
+    /// @param systems systems to run the model on
+    /// @param selected_atoms optional selection of atoms to compute outputs
+    ///     for, or `nullptr` to use all atoms
+    /// @param requested_outputs outputs the model should compute, one per
+    ///     requested output
+    /// @param check_consistency if `true`, run additional checks on the inputs
+    ///     and on the data produced by the model
+    /// @return the computed outputs, one tensor map per requested output
+    inline std::vector<metatensor::TensorMap> execute_model(
+        ExternalModel& model,
+        const std::vector<System>& systems,
+        const metatensor::Labels* selected_atoms,
+        const std::vector<Quantity>& requested_outputs,
+        bool check_consistency
+    ) {
+        std::vector<const mta_system_t*> systems_ptrs;
+        systems_ptrs.reserve(systems.size());
+        for (const auto& system: systems) {
+            systems_ptrs.push_back(system.as_mta_system_t());
+        }
+
+        const mts_labels_t* selected_atoms_ptr = nullptr;
+        if (selected_atoms != nullptr) {
+            selected_atoms_ptr = selected_atoms->as_mts_labels_t();
+        }
+
+        nlohmann::json json = requested_outputs;
+        auto requested_outputs_str = json.dump();
+
+        std::vector<mts_tensormap_t*> outputs(requested_outputs.size(), nullptr);
+
+        auto status = mta_execute_model(
+            *model.as_mta_model_t(),
+            systems_ptrs.data(),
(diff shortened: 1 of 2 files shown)


metatomic-core/include/metatomic/model.hpp
```diff
@@ -356,4 +356,59 @@ namespace metatomic {
...
+    ///     for, or `nullptr` to use all atoms
+    /// @param requested_outputs outputs the model should compute, one per
+    ///     requested output
+    /// @param check_consistency if `true`, run additional checks on the inputs
+    ///     and on the data produced by the model
+    /// @return the computed outputs, one tensor map per requested output
+    inline std::vector<metatensor::TensorMap> execute_model(
+        ExternalModel& model,
+        const std::vector<System>& systems,
+        const metatensor::Labels* selected_atoms,
```
Luthaf: This should be an std::optional, don't you think?

metatomic-core/include/metatomic/model.hpp
```diff
@@ -356,4 +356,59 @@ namespace metatomic {
...
+    /// @param systems systems to run the model on
+    /// @param selected_atoms optional selection of atoms to compute outputs
+    ///     for, or `nullptr` to use all atoms
+    /// @param requested_outputs outputs the model should compute, one per
+    ///     requested output
+    /// @param check_consistency if `true`, run additional checks on the inputs
+    ///     and on the data produced by the model
+    /// @return the computed outputs, one tensor map per requested output
+    inline std::vector<metatensor::TensorMap> execute_model(
+        ExternalModel& model,
```
Luthaf: Should this be a `BaseModel&`?

metatomic-core/include/metatomic/model.hpp
RMeli: Possibly? I do not particularly like optionals in the middle of function signature. Maybe we should swap it with `requested_outputs` and default both `selected_atoms=std::nullopt` and `check_consistency=true`?

Luthaf: >  I do not particularly like optionals in the middle of function signature.

Why? I've never been bothered by these, so genuinely curious.

I'd rather keep the same parameter order across all languages, so if we swap here we should swap everywhere. IMO `check_consistency` should default to `false`, and the user should make a choice for `selected_atoms`

RMeli: Personal preference, I guess. Mostly because I've seen a lot of `{}` instead of `std::nullopt` being used, which makes things less readable IMO. But alright, let's go with `std::optional` here which probably makes more sense.

metatomic-core/include/metatomic/model.hpp
RMeli: Right, I stated with `ExternalModel` and forgot to extend it. I refactored things a bit to work directly with `BaseModel`.

metatomic-core/include/metatomic/model.hpp
```diff
@@ -72,6 +72,26 @@ namespace metatomic {
...
+        ///     a plain pointer to it and does nothing to keep it alive. It is
+        ///     the caller's responsibility to ensure `model` outlives every use
+        ///     of the returned `mta_model_t`, and to never pass the result to
+        ///     an API that takes ownership of the model (i.e. one that would
+        ///     call `unload`).
+        ///     `BaseModel::to_mta_model` whenever ownership can be transferred.
+        ///
+        /// @param model model to borrow
+        /// @return a non-owning `mta_model_t` view of `model`
+        static mta_model_t borrow_mta_model(BaseModel& model);
```
Luthaf: The code is good, I'm not sure about naming. Maybe `mta_model_view` or something else with `view` in the name? Feels to me that borrow is very Rust-brained and that C++ uses view a lot more, feel free to disagree though

Luthaf (approved):
Thanks @RMeli! Looks like github action is having a bad day, I'll restart and merge once CI is happy

PR #324, opened 2026-09-11
C declaration generation for Python and minimal library loading

Luthaf:
This PR adds ctypes declaration for all C API functions, and port over the library loading + error checking code from metatensor, since the API is identical.

TODO:

- [x] add test for external library loading (which will be used in conda packaging)
- [x] (maybe to be seen when we do the Python plugin) when loading the python package from an engine, we need to re-use the already loaded library

# Reviewer checklist


python/metatomic_core/src/metatomic/_status.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def check_status(status):
+    if status == mta_status_t.MTA_SUCCESS:
+        return
+    else:
+        raise _get_exception(status)
+
+
+def check_pointer(pointer):
```
EricBoittier, line 28: should we do a type check? or is this usually ok?

python/metatomic_core/tests/utils.py
```diff
@@ -0,0 +1,15 @@
...
+
+import metatomic as mta
+
+
+def test_cmake_prefix_path():
+    assert os.path.exists(mta.utils.cmake_prefix_path)
+
+
+def test_library_loading():
+    # temporary test to be removed as soon as some other test actually load the library
```
EricBoittier, line 11: add a TODO?

EricBoittier (commented):
LGTM, only a couple of harmless comments from my side.

edit: apparently the code coverage is not good enough
    codecov/projectFailing after 1s — 74.41% (target 75.00%)

python/metatomic_core/src/metatomic/_status.py
Luthaf: This is only intended to be used internally (the whole module is private), so I'd say we are fine.

docs/src/core/reference/python/index.rst
```diff
@@ -0,0 +1,11 @@
+.. _python-api-core:
+
+Python API reference
+====================
+
+WIP
```
johannes-spies, line 6: I guess we're fine not having the documentation populated at all at this point?

scripts/update-declarations.py
```diff
@@ -0,0 +1,376 @@
...
+
+# ==================================================================================== #
+#                                       main                                           #
+# ==================================================================================== #
+
+
+def main():
+    data = parse_header(METATOMIC_HEADER)
+
+    targets = sys.argv[1:] if len(sys.argv) > 1 else ["python"]
```
johannes-spies: Why does this `sys.argv` logic exist? It seems like an overly complicated way of generating the Python code only in certain situations ("I only generate the Python code if there is `"python"` in `sys.argv`"). To me, it looks like a case in which this script should not have been called in the first place. Where do the arguments come from?

johannes-spies (commented):
I just have a few nitpicks where I am not quite sure. I didn't check the full parses logic as you probably copied it from existing `metatensor` sources. However, I ran the generation locally and it seems to work, so LGTM!

docs/src/core/reference/python/index.rst
Luthaf: yes, we'll need to do a final documentation pass when most of the API is ready

scripts/update-declarations.py
Luthaf: this is leftover from the metatensor version which can also generate bindings for other languages. I'll remove it for now, we can add it back when we need it.

PR #325, opened 2026-09-15
Create a System class in Python

EricBoittier:

# Reviewer checklist


docs/src/core/reference/python/system.rst
```diff
@@ -3,5 +3,41 @@ System
...
+    )
+
+The arrays returned by :py:attr:`System.types`, :py:attr:`System.positions`,
+:py:attr:`System.cell`, and :py:attr:`System.pbc` are read-only views. They keep
+the underlying data alive even if the original :py:class:`System` is deleted.
+
+Pair lists can be attached to a system using :py:meth:`System.add_pairs`. Each
+pair list is identified by a :py:class:`PairListOptions` object. Arbitrary
+per-system data stored as a :py:class:`metatensor.TensorMap` can be attached
+with :py:meth:`System.add_custom_data`.
```
Luthaf: I'd put this in the docstring of System, this way it is available both in the docs and when checking docstrings directly from Python

metatomic-core/CHANGELOG.md
```diff
@@ -16,3 +16,8 @@ a changelog](https://keepachangelog.com/en/1.1.0/) format. This project follows
 
 #### Removed
 -->
+
+### Added
+
+- Added the Python `metatomic.System` class, including access to system data,
+  pair lists, custom data, and the underlying C API pointer.
```
Luthaf:
```suggestion
```

We don't track individual changelog items when targetting metatomic-core

python/metatomic_core/src/metatomic/_system.py
```diff
@@ -311,3 +324,266 @@ def __ge__(self, other) -> bool:
 
     def __hash__(self) -> int:
         return hash(self._comparison_key())
+
+
+def _as_dlpack(array) -> ctypes.POINTER(DLManagedTensorVersioned):
```
Luthaf: I would not limit the code here to numpy. I'm not sure we even need a helper here, we can just call  `array_as_dlpack` (no need to make the array contiguous)

python/metatomic_core/src/metatomic/_system.py
```diff
@@ -311,3 +324,266 @@ def __ge__(self, other) -> bool:
...
+    """Convert a numpy array to a DLPack pointer. The array must outlive the call."""
+    return array_as_dlpack(np.ascontiguousarray(array))
+
+
+def _from_dlpack(tensor) -> np.ndarray:
```
Luthaf: Maybe this could also return the data as other array types? We want the arrays to be torch arrays when working with torch models, numpy when working with numpy models, …

Not sure what the best user-facing API would be though, maybe something like

```
class System
    def __init__(self, ..., arrays_backend):
        ...
        self._arrays_backend = arrays_backend

    def set_arrays_backend(self, backend):
        ...
    
    def positions(self):
        if self._arrays_backend is None:
            raise ValueError("Arrays backend not initialized, please set it")

        if self._arrays_backend == "dlpack":
            # useful for anything not supported below
            return DLPackArray(self._positions)
        elif self._arrays_backend == "numpy":
            return np.array(DLPackArray(self._positions))
        elif self._arrays_backend == "torch":
            return torch.tensor(DLPackArray(self._positions))
        elif self._arrays_backend == "jax":
            return jnp.array(DLPackArray(self._positions))
        else:
            raise ValueError(f"Unknown arrays backend: {self._arrays_backend}")


# auto-guess backend when creating from Python
system = System(positions=np.array(...))
assert system._arrays_backend == "numpy"

# force explicit backend when wrapping external systems 
system = System.from_mta_system(ptr)
system.set_arrays_backend("torch")
```

What do you think?

python/metatomic_core/src/metatomic/_system.py
```diff
@@ -311,3 +324,266 @@ def __ge__(self, other) -> bool:
...
+    """Take a borrowed DLPack tensor from the C API and wrap it as numpy."""
+    return np.from_dlpack(DLPackArray(tensor))
+
+
+def _string_from_mta(printed) -> str:
+    lib = _get_library()
+    try:
+        return lib.mta_string_view(printed).decode("utf8")
+    finally:
+        lib.mta_string_free(printed)
```
Luthaf: worth putting in utils

python/metatomic_core/src/metatomic/_system.py
```diff
@@ -311,3 +324,266 @@ def __ge__(self, other) -> bool:
...
+
+
+class System:
+    """
+    An atomistic system used as input to metatomic models.
+
+    This is a RAII wrapper around the ``mta_system_t`` type from the C API. It
```
Luthaf: RAII is not a Python term, it's a C++ one

python/metatomic_core/src/metatomic/_system.py
```diff
@@ -311,3 +324,266 @@ def __ge__(self, other) -> bool:
...
+        return obj
+
+    def as_mta_system_t(self):
+        """
+        Get the underlying C pointer for this :py:class:`System`.
+
+        This class still manages the system memory after the call. Use
+        :py:meth:`System.release` to take ownership of the pointer.
+        """
+        self._check_ptr()
```
Luthaf: instead of making this a separate function one needs to remember to call, can we make `_ptr` a property that does the check for us?

Luthaf: Or better, do the check in this function, and have everyone go through it instead of directly accessing self._ptr

python/metatomic_core/src/metatomic/_system.py
```diff
@@ -311,3 +324,266 @@ def __ge__(self, other) -> bool:
...
+
+    @property
+    def pbc(self) -> np.ndarray:
+        """Periodic boundary conditions, as an array with shape ``(3,)``."""
+        return self._data(mta_system_data_kind.MTA_SYSTEM_DATA_PBC)
+
+    def _options_json(self, options: Union[PairListOptions, str]) -> bytes:
+        if isinstance(options, PairListOptions):
+            return json.dumps(options.to_dict()).encode("utf8")
+        return str(options).encode("utf8")
```
Luthaf: This should be a free function or directly inlined where we use it. there is no point making it a member function

python/metatomic_core/tests/system.py
```diff
@@ -208,3 +211,170 @@ def without(key):
...
         with pytest.raises(ValueError, match=message):
             PairListOptions.from_dict(data)
+
+
+### ================================================================================ ###
+###                                      System                                      ###
+### ================================================================================ ###
+
+
+def _test_system(n_atoms=4, positions_dtype=np.float64):
```
Luthaf: can you make this & the other similar functions below a pytest fixture?

python/metatomic_core/src/metatomic/_system.py
EricBoittier: > Maybe this could also return the data as other array types? We want the arrays to be torch arrays when working with torch models, numpy when working with numpy models, …
> 
> Not sure what the best user-facing API would be though, maybe something like
> 
> ```
> class System
>     def __init__(self, ..., arrays_backend):
>         ...
>         self._arrays_backend = arrays_backend
> 
>     def set_arrays_backend(self, backend):
>         ...
>     
>     def positions(self):
>         if self._arrays_backend is None:
>             raise ValueError("Arrays backend not initialized, please set it")
> 
>         if self._arrays_backend == "dlpack":
>             # useful for anything not supported below
>             return DLPackArray(self._positions)
>         elif self._arrays_backend == "numpy":
>             return np.array(DLPackArray(self._positions))
>         elif self._arrays_backend == "torch":
>             return torch.tensor(DLPackArray(self._positions))
>         elif self._arrays_backend == "jax":
>             return jnp.array(DLPackArray(self._positions))
>         else:
>             raise ValueError(f"Unknown arrays backend: {self._arrays_backend}")
> 
> 
> # auto-guess backend when creating from Python
> system = System(positions=np.array(...))
> assert system._arrays_backend == "numpy"
> 
> # force explicit backend when wrapping external systems 
> system = System.from_mta_system(ptr)
> system.set_arrays_backend("torch")
> ```
> 
> What do you think?

it looks solid to me... would you allow:
def positions(self, as_array_backend=None):

if as_array_backend=None --> default behaviour

or maybe something like the numpy api, giving the method, arrays.to("array_backend_enum")?
also we should think about dtypes -- probably these need to be given + sensible defaults (e.g. f32 flavoured for vanilla torch/jax)

python/metatomic_core/src/metatomic/_system.py
EricBoittier: fair

python/metatomic_core/src/metatomic/_system.py
EricBoittier: that's logical

python/metatomic_core/src/metatomic/_system.py
Luthaf: >  dtypes -- probably these need to be given + sensible defaults (e.g. f32 flavoured for vanilla torch/jax)

This is handled by DLPack, as well as device!

EricBoittier (commented):
👀

python/metatomic_core/tests/system.py
```diff
@@ -374,7 +385,109 @@ def test_system_ownership():
...
+    assert system.arrays_backend == "torch"
+    assert isinstance(system.types, torch.Tensor)
+    assert isinstance(system.positions, torch.Tensor)
+    assert isinstance(system.cell, torch.Tensor)
+    assert isinstance(system.pbc, torch.Tensor)
+    assert system.positions[3, 0] == 10.0
+
+
+def test_system_set_arrays_backend_torch(system):
+    torch = pytest.importorskip("torch")
```
Luthaf, line 494: are these tested on CI?

python/metatomic_core/tests/system.py
```diff
@@ -374,7 +385,109 @@ def test_system_ownership():
...
+
+    system.set_arrays_backend("torch")
+    assert system.arrays_backend == "torch"
+    assert isinstance(system.positions, torch.Tensor)
+    assert system.positions[3, 0] == 10.0
+
+
+def test_system_arrays_backend_jax():
+    jax = pytest.importorskip("jax")
```
Luthaf, line 519: Same, do we install jax on CI?

python/metatomic_core/src/metatomic/_system.py
```diff
@@ -461,50 +521,103 @@ def __len__(self) -> int:
...
-        self._check_ptr()
         unit = mta_string_t()
-        self._lib.mta_system_get_length_unit(self._ptr, ctypes.byref(unit))
+        self._lib.mta_system_get_length_unit(self.as_mta_system_t(), ctypes.byref(unit))
         return _string_from_mta(unit)
 
-    def _data(self, kind) -> np.ndarray:
-        self._check_ptr()
+    @property
+    def arrays_backend(self):
```
Luthaf:
```suggestion
    def arrays_backend(self) -> str:
```

python/metatomic_core/tests/system.py
```diff
@@ -218,52 +218,63 @@ def without(key):
...
-                "cell_shift_c",
-            ],
-            np.array([[0, 1, 0, 0, 0]], dtype=np.int32),
-        ),
-        components=[Labels("xyz", np.array([[0], [1], [2]], dtype=np.int32))],
-        properties=Labels("distance", np.array([[0]], dtype=np.int32)),
-    )
+@pytest.fixture
+def pair_block():
+    def make():
```
Luthaf: why are you returning a function?

python/metatomic_core/tests/system.py
```diff
@@ -218,52 +218,63 @@ def without(key):
...
-    block = TensorBlock(
-        values=np.array([[42.0]], dtype=np.float64),
-        samples=Labels("sample", np.array([[0]], dtype=np.int32)),
-        components=[],
-        properties=Labels("property", np.array([[0]], dtype=np.int32)),
-    )
-    return TensorMap(Labels("key", np.array([[0]], dtype=np.int32)), [block])
+@pytest.fixture
+def custom_data():
+    def make():
```
Luthaf: same

python/metatomic_core/tests/system.py
```diff
@@ -306,12 +315,11 @@ def test_system_data():
...
 
 
-def test_system_pairs():
-    system = _test_system(4)
+def test_system_pairs(system, pair_block):
     options = PairListOptions(
         cutoff=1.0, full_list=True, strict=False, requestors=["test"]
     )
-    system.add_pairs(options, _pair_block())
+    system.add_pairs(options, pair_block())
```
Luthaf: instead of making the fixture a function, you can do something like `pair_block.copy()` here

Luthaf: (not 100% sure on the function name: copy/clone/…)

Luthaf (commented):
Did you check for compatibility with `torch.compile`/`jax.jit`?

EricBoittier:
> Did you check for compatibility with `torch.compile`/`jax.jit`?

I generated a notebook to answer these questions (attached as PDF, sorry for LLM-braining
[system-compile-jit.pdf](https://github.com/user-attachments/files/32329552/system-compile-jit.pdf)
)

It shows:

- numpy / torch / jax construction and getters work

- extracted arrays work with torch.compile (fullgraph=True) and jax.jit

- constructing System inside torch.compile(..., fullgraph=True) fails on ctypes POINTER.__new__ and ctypes graph-breaks to eager if fullgraph!=True

- jax.jit rejects System; constructing inside jit fails (__dlpack__) since `metatomic.System` is a host-side ctypes wrapper around `mta_system_t`, not a traced model type.

python/metatomic_core/tests/system.py
EricBoittier: > Same, do we install jax on CI?
Yes — but only after 67d8cd23, and only on Linux Python < 3.14.

Luthaf:
I'd say it's fine for now if we can not construct a system inside a `torch.compile`/`jax.jit` bit of code, but it would be nice to be able to compile a function like `squared_sum_positions(system): return torch.sum(system.positions)**2`. Looks like this works for torch, how hard would it be to make it work for jax? We don't have to do it in this PR, just make sure we are not painting ourself in a corner!

docs/src/core/reference/python/system.rst
```diff
@@ -3,5 +3,11 @@ System
...
 .. currentmodule:: metatomic
 
+The following classes store atomistic systems passed to a metatomic model, and
+the options used to request pair lists (neighbor lists) for those systems.
+
 .. autoclass:: PairListOptions
    :members:
+
+.. autoclass:: System
+   :members:
```
Luthaf, line 13: Small nit: System is more important and thus should come first in the API docs

```suggestion
.. autoclass:: System
   :members:

.. autoclass:: PairListOptions
   :members:
```

metatomic-core/include/metatomic.h
```diff
@@ -489,6 +489,9 @@ enum mta_status_t mta_system_get_length_unit(const mta_system_t *system, mta_str
  * This function **takes ownership** of `pairs`. The caller must not use the
  * block after calling this function.
  *
+ * The system must not have outstanding borrowed views from
+ * `mta_system_get_data`. Release those tensors first.
+ *
```
Luthaf, line 494: I don't get this? `mta_system_add_pairs` should not interact with `mta_system_get_data`

metatomic-core/src/c_api/system.rs
```diff
@@ -276,12 +279,11 @@ pub unsafe extern "C" fn mta_system_add_pairs(
 
         let pairs = unsafe { TensorBlock::from_raw(pairs) };
 
-        let mut system = unsafe { mta_system_t::from_raw(system.cast_const()) };
+        let mut system = std::mem::ManuallyDrop::new(unsafe {
```
Luthaf, line 282: `ManuallyDrop` is so we don't drop the system if add_pairs fails?

python/metatomic_core/src/metatomic/_system.py
```diff
@@ -311,3 +309,405 @@ def __ge__(self, other) -> bool:
...
+    def pbc(self):
+        """
+        Periodic boundary conditions, as an array with shape ``(3,)``.
+
+        The returned array uses the configured :py:attr:`arrays_backend` and is
+        a read-only view.
+        """
+        return self._data(mta_system_data_kind.MTA_SYSTEM_DATA_PBC)
+
+    def add_pairs(self, options: Union[PairListOptions, str], pairs: TensorBlock):
```
Luthaf: options should only be PairListOptions here, unless you have a use case for the str case? The idea is that we use JSON at the C boundary, but nobody has to write JSON and/or see that we are using it!

python/metatomic_core/src/metatomic/_system.py
```diff
@@ -311,3 +309,405 @@ def __ge__(self, other) -> bool:
...
+            :py:attr:`positions`, :py:attr:`cell`, and :py:attr:`pbc`. One of
+            ``"numpy"``, ``"torch"``, ``"jax"``, or ``"dlpack"``. Guessed from
+            ``positions`` when omitted.
+        """
+        self._lib = _get_library()
+        self._is_view = False
+        if arrays_backend is None:
+            self._arrays_backend = _guess_arrays_backend(positions)
+        else:
+            self.set_arrays_backend(arrays_backend)
```
Luthaf: should we check that all data has the same array backend? It is not strictly required, but it might help.

python/metatomic_core/src/metatomic/_system.py
```diff
@@ -311,3 +309,405 @@ def __ge__(self, other) -> bool:
...
+        Arrays backend used by :py:attr:`types`, :py:attr:`positions`,
+        :py:attr:`cell`, and :py:attr:`pbc`.
+
+        One of ``"numpy"``, ``"torch"``, ``"jax"``, or ``"dlpack"``, or
+        ``None`` if this :py:class:`System` was created from a C pointer and
+        :py:meth:`set_arrays_backend` has not been called yet.
+        """
+        return self._arrays_backend
+
+    def set_arrays_backend(self, backend: str):
```
Luthaf: if we make `arrays_backend` a property, should this be a setter for this property?

python/metatomic_core/src/metatomic/_system.py
```diff
@@ -311,3 +309,405 @@ def __ge__(self, other) -> bool:
...
+        """
+        return self._data(mta_system_data_kind.MTA_SYSTEM_DATA_TYPES)
+
+    @property
+    def positions(self):
+        """
+        Positions of all atoms, as an array with shape ``(n_atoms, 3)``.
+
+        The returned array uses the configured :py:attr:`arrays_backend` and is
+        a read-only view.
```
Luthaf, line 618: Are we sure these are read-only?

tox.ini
```diff
@@ -58,6 +58,9 @@ deps =
     {[testenv]testing_deps}
     {[testenv]packaging_deps}
     {[testenv]metatomic_deps}
+    numpy
+    torch=={env:METATOMIC_TESTS_TORCH_VERSION:2.13}.*
+    jax; sys_platform == "linux" and python_version < "3.14"
```
Luthaf: why linux only?

Luthaf (commented):
I do not understand the new requirement of releasing all arrays views before adding pairs/custom data. Can you explain?

EricBoittier:
> I do not understand the new requirement of releasing all arrays views before adding pairs/custom data. Can you explain?

The two APIs do not share buffers. 
add_pairs only does system_mut.pairs.insert(...) ([system.rs](metatomic-core/src/system.rs) 371).
They're coupled because mta_system_get_data clones the Arc<System> into manager_ctx ([c_api/system.rs](metatomic-core/src/c_api/system.rs) 191–209). add_pairs / add_custom_data go through Arc::get_mut (281–285 / 408–412), so any live getter view makes strong_count > 1, so mutation fails 

We can keep this, or allow adds with live views (interior mutability on pairs / custom_data only). Which do you want?

On the topic of mutability, jax arrays strictly immutable. In torch, arrays may still look writable depending on version; the C comment on get_data is still “do not modify” ([c_api/system.rs](metatomic-core/src/c_api/system.rs) 168–170).
I would go for immutable, even if pytorch doesn't strictly require it.

Luthaf:
Ah, right there is the `Arc::get_mut` thing. Ok, makes sense then!


> On the topic of mutability, jax arrays strictly immutable. [...]

This should already be handled by DLPack, which has a flag for immutable data. So unless we gain something by specifying it ourself, I'd leave read-only considerations to DLPack and not mention it ourself.

metatomic-core/tests/system.cpp
```diff
@@ -482,6 +482,62 @@ TEST_CASE("system custom data") {
     mta_system_free(system);
 }
 
+TEST_CASE("system additions keep caller ownership on error") {
```
Luthaf: not clear what exactly this is testing

Luthaf:
```suggestion
TEST_CASE("system isn't dropped on failure") {
```

or anything better

Luthaf: the function body then needs comments about what's going on and why

python/metatomic_core/src/metatomic/_system.py
```diff
@@ -311,3 +309,413 @@ def __ge__(self, other) -> bool:
...
+    """Guess the arrays backend from a constructor argument, typically positions."""
+    if isinstance(array, np.ndarray):
+        return "numpy"
+
+    module = type(array).__module__
+    name = type(array).__name__
+    if module.startswith("torch") and name == "Tensor":
+        return "torch"
+    if module.startswith("jax") or module.startswith("jaxlib"):
+        return "jax"
```
Luthaf: should maybe use isinstance instead since we could have subclasses of these. See https://github.com/metatensor/metatensor/blob/9dfa936675f8ee92087ca63d122ac04f76333447/python/metatensor_operations/metatensor_operations/_dispatch.py#L44

python/metatomic_core/src/metatomic/_system.py
```diff
@@ -311,3 +309,413 @@ def __ge__(self, other) -> bool:
...
+    if backend == "dlpack":
+        return wrapped
+    if backend == "numpy":
+        return np.from_dlpack(wrapped)
+    if backend == "torch":
+        import torch
+
+        return torch.from_dlpack(wrapped)
+    if backend == "jax":
+        import jax.numpy as jnp
```
Luthaf: this works with CUDA arrays as well, right?

python/metatomic_core/src/metatomic/_system.py
```diff
@@ -311,3 +309,413 @@ def __ge__(self, other) -> bool:
...
+
+        return jnp.from_dlpack(wrapped)
+
+    raise ValueError(f"Unknown arrays backend: {backend}")
+
+
+def _pair_options_json(options: PairListOptions) -> bytes:
```
Luthaf: not worth a helper IMO

python/metatomic_core/src/metatomic/_system.py
```diff
@@ -311,3 +309,413 @@ def __ge__(self, other) -> bool:
...
+        check_pointer(tensor)
+        return _array_from_dlpack(tensor, self._arrays_backend)
+
+    @property
+    def types(self):
+        """
+        Atomic types of all atoms, as an array with shape ``(n_atoms,)``.
+
+        The returned array uses the configured :py:attr:`arrays_backend` and is
+        a read-only view.
```
Luthaf:
```suggestion
        The returned array uses the configured :py:attr:`arrays_backend`.
```

and same below

python/metatomic_core/src/metatomic/_system.py
```diff
@@ -311,3 +309,413 @@ def __ge__(self, other) -> bool:
...
+        """
+        return self._data(mta_system_data_kind.MTA_SYSTEM_DATA_PBC)
+
+    def add_pairs(self, options: PairListOptions, pairs: TensorBlock):
+        """
+        Add a pair list (neighbor list) to this system.
+
+        Ownership of ``pairs`` is transferred to this :py:class:`System`.
+        All arrays previously returned by :py:attr:`types`, :py:attr:`positions`,
+        :py:attr:`cell`, or :py:attr:`pbc` must be released first.
```
Luthaf:
```suggestion
        Add a pair list (neighbor list) to this system.

        Ownership of ``pairs`` is transferred to this :py:class:`System`, the pairs can no longer be used after this function.
        
        This function should be called first, before anyone tries to access  :py:attr:`types`, :py:attr:`positions`, :py:attr:`cell`, or :py:attr:`pbc`.
```

Luthaf: same below

python/metatomic_core/src/metatomic/_system.py
```diff
@@ -311,3 +309,413 @@ def __ge__(self, other) -> bool:
...
+        self._lib.mta_system_add_pairs(
+            self.as_mta_system_t(), _pair_options_json(options), pairs.release()
+        )
+
+    def pairs(self, options: PairListOptions) -> TensorBlock:
+        """
+        Get a previously stored pair list matching ``options``.
+
+        The returned block is a non-owning view into data owned by this
+        :py:class:`System`.
```
Luthaf:
```suggestion
```

python/metatomic_core/src/metatomic/_system.py
```diff
@@ -311,3 +309,413 @@ def __ge__(self, other) -> bool:
...
+        self._lib.mta_system_add_custom_data(
+            self.as_mta_system_t(), str(name).encode("utf8"), data.release()
+        )
+
+    def custom_data(self, name: str) -> TensorMap:
+        """
+        Get the custom data previously stored under ``name``.
+
+        The returned tensor map is a non-owning view into data owned by this
+        :py:class:`System`.
```
Luthaf:
```suggestion
```

python/metatomic_core/tests/system.py
```diff
@@ -68,6 +70,15 @@ def test_pair_options_requestors(pair_options):
     options.requestors = ["a", "", "b", "a"]
     assert options.requestors == ["a", "b"]
 
+    # a bare string is a Sequence[str] but would silently split into characters
+    options.requestors = ["keep-me"]
+    with pytest.raises(TypeError, match="sequence of strings, not a single string"):
```
Luthaf: personal preference, but I'd rather have `pytest.raises` check the full error message instead of a substring, which is typically easier to write as 
```suggestion
    message = "znjnzrgoguhiqo sequence of strings, not a single string"
    with pytest.raises(TypeError, match=message):
```

python/metatomic_core/tests/system.py
```diff
@@ -208,3 +219,351 @@ def without(key):
...
+    assert system.arrays_backend == "numpy"
+    assert isinstance(system.positions, np.ndarray)
+
+
+def test_system_construction_errors():
+    types = np.array([1, 2, 3], dtype=np.float32)
+    positions = np.zeros((3, 3), dtype=np.float32)
+    cell = np.eye(3, dtype=np.float32)
+    pbc = np.array([True, True, True])
+    with pytest.raises(MetatomicError, match="types"):
```
Luthaf: same with the error message (and everywhere in this file)

python/metatomic_core/tests/system.py
```diff
@@ -208,3 +219,351 @@ def without(key):
...
+
+    pbc = np.asarray(system.pbc)
+    assert pbc.shape == (3,)
+    assert not pbc.flags.writeable
+    assert bool(pbc[0]) is True
+    assert bool(pbc[1]) is False
+    assert bool(pbc[2]) is True
+
+    # DLPack views keep the backing system storage alive.
+    del system
```
Luthaf: this doesn't do what you think it does =)

system is a fixture so there are already multiple references to it.

See https://github.com/metatensor/metatensor/blob/9dfa936675f8ee92087ca63d122ac04f76333447/python/metatensor_core/tests/tensor.py#L107-L134

and https://github.com/metatensor/metatensor/blob/9dfa936675f8ee92087ca63d122ac04f76333447/python/metatensor_core/tests/data.py#L100-L143 for example checking refcounts

python/metatomic_core/tests/system.py
```diff
@@ -208,3 +219,351 @@ def without(key):
...
+
+
+def test_system_pairs(system, pair_block):
+    options = PairListOptions(
+        cutoff=1.0, full_list=True, strict=False, requestors=["test"]
+    )
+    system.add_pairs(options, pair_block.copy())
+
+    other = PairListOptions(cutoff=22.3, full_list=False)
+    system.add_pairs(other, pair_block.copy())
```
Luthaf: maybe remove the copy and assert the pair_block is now empty? (not sure if we have a public API for this though)

python/metatomic_core/tests/system.py
```diff
@@ -208,3 +219,351 @@ def without(key):
...
+
+
+def test_system_missing_pairs_is_an_error():
+    types = np.array([1, 4], dtype=np.int32)
+    positions = np.array([[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]], dtype=np.float64)
+    cell = np.zeros((3, 3), dtype=np.float64)
+    pbc = np.array([False, False, False])
+    system = System("nm", types, positions, cell, pbc)
+    missing = PairListOptions(cutoff=9.0, full_list=False)
+    with pytest.raises(MetatomicError):
```
Luthaf: check the message

python/metatomic_core/tests/system.py
```diff
@@ -208,3 +219,351 @@ def without(key):
...
+    assert system.size == 4
+
+    del positions
+    system.add_pairs(options, pair_block.copy())
+    system.add_custom_data("test::my_data", custom_data.copy())
+    assert len(system.known_pairs()) == 1
+    assert system.known_custom_data() == ["test::my_data"]
+
+
+def test_system_failed_additions_keep_ownership(system, pair_block, custom_data):
```
Luthaf: not sure what this is testing on the Python side

python/metatomic_core/tests/system.py
```diff
@@ -208,3 +219,351 @@ def without(key):
...
+    system.add_pairs(PairListOptions(cutoff=1.0, full_list=True), pair_block.copy())
+    assert len(system.known_pairs()) == 1
+
+    view.arrays_backend = "numpy"
+    assert view.arrays_backend == "numpy"
+    assert isinstance(view.positions, np.ndarray)
+    assert view.positions[3, 0] == 10.0
+
+
+def test_system_arrays_backend_unknown(system):
```
Luthaf: can be folded in the test above, up to you

python/metatomic_core/tests/system.py
```diff
@@ -208,3 +219,351 @@ def without(key):
...
+    )
+    pbc = torch.tensor([True, False, True])
+
+    system = System("nm", types, positions, cell, pbc)
+    assert system.arrays_backend == "torch"
+    assert isinstance(system.types, torch.Tensor)
+    assert isinstance(system.positions, torch.Tensor)
+    assert isinstance(system.cell, torch.Tensor)
+    assert isinstance(system.pbc, torch.Tensor)
+    assert system.positions[3, 0] == 10.0
```
Luthaf: do we also want tests on other devices here?

python/metatomic_core/tests/system.py
```diff
@@ -208,3 +219,351 @@ def without(key):
...
+        dtype=torch.float64,
+    )
+    pbc = torch.tensor([True, False, True])
+
+    system = System("nm", types, positions, cell, pbc)
+    assert system.arrays_backend == "torch"
+    assert isinstance(system.types, torch.Tensor)
+    assert isinstance(system.positions, torch.Tensor)
+    assert isinstance(system.cell, torch.Tensor)
+    assert isinstance(system.pbc, torch.Tensor)
```
Luthaf: Let's also do a test creating a system with torch tensor and then setting array backend to numpy

python/metatomic_core/tests/system.py
```diff
@@ -208,3 +219,351 @@ def without(key):
...
+def test_system_arrays_backend_set_torch(system):
+    torch = pytest.importorskip("torch")
+
+    system.arrays_backend = "torch"
+    assert system.arrays_backend == "torch"
+    assert isinstance(system.positions, torch.Tensor)
+    assert system.positions[3, 0] == 10.0
+
+
+def test_system_mixed_arrays_backend_requires_explicit():
```
Luthaf: test name does not match the content

python/metatomic_core/tests/system.py
```diff
@@ -208,3 +219,351 @@ def without(key):
...
+    with pytest.raises(ValueError, match="same arrays backend"):
+        System("nm", types, positions, cell, pbc)
+
+    system = System("nm", types, positions, cell, pbc, arrays_backend="torch")
+    assert system.arrays_backend == "torch"
+    assert isinstance(system.positions, torch.Tensor)
+    assert system.positions[3, 0] == 10.0
+
+
+def test_system_torch_compile_squared_sum_positions(system):
```
Luthaf:
```suggestion
def test_system_torch_compile(system):
```

python/metatomic_core/tests/system.py
```diff
@@ -208,3 +219,351 @@ def without(key):
...
+    torch = pytest.importorskip("torch")
+    system.arrays_backend = "torch"
+
+    def squared_sum_positions(positions):
+        return torch.sum(positions) ** 2
+
+    positions = system.positions
+    expected = squared_sum_positions(positions)
+    compiled = torch.compile(squared_sum_positions, backend="eager")
+    torch.testing.assert_close(compiled(positions), expected)
```
Luthaf: this is not very useful, if 

```
def squared_sum_positions(system):
        return torch.sum(system.positions) ** 2
```

does not work, we should remove this test

tox.ini
```diff
@@ -58,6 +58,9 @@ deps =
     {[testenv]testing_deps}
     {[testenv]packaging_deps}
     {[testenv]metatomic_deps}
+    numpy
+    torch=={env:METATOMIC_TESTS_TORCH_VERSION:2.13}.*
+    jax; python_version < "3.14"
```
Luthaf:
```suggestion
    jax
```

Does not seems necessary

PR #328, opened 2026-09-15
Fix Metal atomic type buffer offsets

EricBoittier:
The Metal atomic-type validation wrapped DLPack memory at its page-aligned base, but both the GPU binding and CPU fallback readback ignored the tensor offset within that page. As a result, the kernel could inspect unrelated memory and report nondeterministic invalid atomic types.

This uses the offset recorded by MetalBufferRef for both operations, matching the other Metal kernels.

Tests:
- cargo fmt --all -- --check
- check_atomic_types_kernel passed 5 consecutive runs
- cargo test -p metatomic-core: 240 Rust unit tests, 4 install tests, and 11 C/C++ tests passed

LLM disclosure: The diagnosis, code change, commit message, and pull-request description were generated by OpenAI Codex, an LLM. The contributor explicitly requested this disclosure.

- [x] Tests updated (the existing regression test failed before this fix and passes after it)
- [ ] Documentation updated (not applicable; no public behavior or API change)
- [ ] Issue referenced (no issue)

# Reviewer checklist

- [ ] CHANGELOG updated with public API or any other important changes (not applicable)


EricBoittier:
Closing for now so the branch and PR text can be reviewed before submission.

Luthaf:
I have an error in this function when running tests locally as well, not sure why since the code worked last time I checked 🤷. I'll let you finish cleaning up the fix!

PR #331, opened 2026-09-15
C++ plugin registration

RMeli:
Implement C++ plugin registration with the `MTA_REGISTER_CXX_PLUGIN` macro.

The code in 1de850a52f9a18a4cc2686ccbe89b967d9eab92c was suggested by Kimi-2.7-Code during a code review and implemented using Opus 5, so that a throwing C++ plugin's exception now reaches the caller with its original type (something I originally overlooked).

[📚 Download documentation for this pull-request](https://nightly.link/metatensor/metatomic/actions/artifacts/10498000059.zip)

[⚙️ Download Python wheels for this pull-request (you can install these with pip)](https://nightly.link/metatensor/metatomic/actions/artifacts/10498058560.zip)


Luthaf:
> The code in https://github.com/metatensor/metatomic/commit/1de850a52f9a18a4cc2686ccbe89b967d9eab92c was suggested by Kimi-2.7-Code during a code review and implemented using Opus 5, so that a throwing C++ plugin's exception now reaches the caller with its original type (something I originally overlooked).

I don't think we need a custom status code for this, it works without one in the C++ API of metatensor. (but this could also be thanks to the different treatment of `CallbackError` in here vs metatensor, I'll triple check).

EricBoittier (commented):
LGTM

Luthaf:
I did a bit of refactoring here while splitting the changes into two separate commits, mainly making extra parameters `std::optional` in load_model; and re-using the C macro for plugins inside the C++ one, to make sure we don't have to maintain both places.

Let me know if the changes are good for you, then we can merge!

RMeli:
I can't self-approve, but your changes LGTM, thanks!

